PR #3, opened 2021-02-23
Add input validation, minor refactor

rekmarks:
This PR adds input validation to ensure that only properly formatted paths are processed by `deriveKeyFromPath`. Aside from internal import/export refactoring, the only functional change is the added input validation, and the tests, which are functionally unchanged, pass.

This PR is intended as the first entry before a major version bump, due to this package's anticipated production use. Therefore, the following breaking, but not cryptographically relevant changes, are made:
- `deriveKeyFromPath`: Change the order of parameters
  - The first parameter was often `null`, which is unergonomic
- Add `index.js` file to `derivers` directory, re-export individual derivers from there
  - We (I?) love being agnostic of directory structure when importing!
- Update readme
- Delete dead code

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,48 +1,73 @@
-// node
-const crypto = require('crypto')
-const assert = require('assert')
-// npm
 const bip39 = require('bip39')
-const secp256k1 = require('secp256k1')
-const createKeccakHash = require('keccak')
 
-const derivers = {
-  'bip32': require('./derivers/bip32'),
-  'bip39': require('./derivers/bip39'),
-}
+const derivers = require('./derivers')
 
 module.exports = {
-  mnemonicToSeed,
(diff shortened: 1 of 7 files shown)


src/index.js
```diff
@@ -1,48 +1,91 @@
...
+
+  // BIP-32 segments cannot initiate HD paths
+  if (!pathSegment.startsWith('bip39') && !parentKey) {
+    throw new Error('Must specify parent key if the first path of the path segment is not BIP-39.')
+  }
+
+  // The parent key must be a Buffer
+  if (parentKey && !Buffer.isBuffer(parentKey)) {
+    throw new Error('Parent key must be a Buffer if specified.')
+  }
```
rekmarks, line 69: This here is the crux of this review.

PR #14, opened 2021-02-27
Fix deriveKeyFromPath interface, refactor derivers

rekmarks:
Ensures that `deriveKeyFromPath` `parentKey` parameter is optional. This required some refactoring of the internal derivers, since BIP-32 derivation requires a `parentKey` while BIP-39 does not. The public interface of this package remains unchanged.


astarinmymind (approved):
Nice n Simple!

PR #15, opened 2021-02-27
Ensure bip39 validation rejects non-lowercase seed phrases

rekmarks:
Fixes a bug in input validation where BIP-39 seed phrases could contain non-lowercase letters.


astarinmymind (approved):
yay regex!!

PR #37, opened 2022-03-09
Allow deriving hardened `change` and `address_index`

### LINKED ISSUE #34: Let the user configure hardening during address key derivation
Recall that a BIP-44 path looks like this: `m / purpose' / coin_type' / account' / change / address_index`

`purpose` and `coin_type` nodes are always hardened, but deeper nodes may not be for all chains and forms of key derication. `getBIP44AddressKeyDeriver` and `deriveBIP44AddressKey` should default hardening behavior for `coin_type` descendants as specified by BIP-44, but we should make the behavior configurable so that `change` and `address_index` nodes can be hardened if desired by the user.

For example, Solana uses hardened derivation for all nodes along the derivation path.

Mrtenz:
This makes it possible to use a hardened `change` and `address_index`.

- Changed types to allow both hardened and unhardened indices for these paths.
- Changed some validation functions to allow both hardened and unhardened indices for these paths.
- Added an option `isHardened` to the deriver function, to derive a hardened `address_index`.

Closes #34.

### DIFF
--- a/src/BIP44CoinTypeNode.test.ts
+++ b/src/BIP44CoinTypeNode.test.ts
@@ -578,6 +578,8 @@ describe('deriveBIP44AddressKey', () => {
     [
       {},
       { change: {} },
+      { change: null },
+      { change: undefined },
       { address_index: {} },
       { address_index: 'foo' },
       { address_index: { index: 1 } },
--- a/src/BIP44CoinTypeNode.test.ts
+++ b/src/BIP44CoinTypeNode.test.ts
@@ -1,5 +1,4 @@
 import fixtures from '../test/fixtures';
-import { BIP_32_HARDENED_OFFSET } from './constants';
 import {
   BIP_44_COIN_TYPE_DEPTH,
   BIP44Node,
@@ -357,6 +356,35 @@ describe('deriveBIP44AddressKey', () => {
     ).toStrictEqual(expectedKey);
   });
 
+  it('derives an address_index key (default inputs, object address_index)', () => {
+    const coinType = 60;
+    const parentNode = new BIP44CoinTypeNode([
+      defaultBip39NodeToken,
+      BIP44PurposeNodeToken,
+      `bip32:${coinType}'`,
+    ]);
+
+    const expectedKey = new BIP44Node({
+      derivationPath: [
+        defaultBip39NodeToken,
+        BIP44PurposeNodeToken,
+        `bip32:${coinType}'`,
+        `bip32:0'`,
+        `bip32:0`,
+        `bip32:3333`,
+      ],
+    }).key;
+
+    expect(
+      deriveBIP44AddressKey(parentNode, {
+        address_index: {
+          index: 3333,
+          hardened: false,
+        },
+      }).toString('base64'),
+    ).toStrictEqual(expectedKey);
+  });
+
   it('derives an address_index key (default inputs, hardened address_index)', () => {
     const coinType = 60;
     const parentNode = new BIP44CoinTypeNode([
@@ -378,7 +406,10 @@ describe('deriveBIP44AddressKey', () => {
 
     expect(
       deriveBIP44AddressKey(parentNode, {
-        address_index: BIP_32_HARDENED_OFFSET + 3333,
+        address_index: {
+          index: 3333,
+          hardened: true,
+        },
       }).toString('base64'),
     ).toStrictEqual(expectedKey);
   });
@@ -524,10 +555,8 @@ describe('deriveBIP44AddressKey', () => {
     ]);
 
     [
-      {},
       { account: -1 },
       { change: 1.1 },
-      { address_index: 'foo' },
       { account: NaN },
       { account: null },
       { account: {} },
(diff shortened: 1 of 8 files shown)


src/BIP44CoinTypeNode.test.ts
```diff
@@ -356,6 +357,32 @@ describe('deriveBIP44AddressKey', () => {
...
+        `bip32:${coinType}'`,
+        `bip32:0'`,
+        `bip32:0`,
+        `bip32:3333'`,
+      ],
+    }).key;
+
+    expect(
+      deriveBIP44AddressKey(parentNode, {
+        address_index: BIP_32_HARDENED_OFFSET + 3333,
```
Mrtenz, line 381: Alternatively, instead of simply adding the hardened offset to the `address_index` (or `change`), we could also:

- Allow specifying a string, like `3333'`
- Allow specifying an object, like `{ index: number; isHardened: boolean; }`

FrederikBolding: I would lean towards specifying an object. At least that matches the current API

FrederikBolding: But perhaps we need an object for each index then? `deriveBIP44AddressKey(parentNode, { index: 60, isHardened: true }, (...))`

Mrtenz: Yeah, I was thinking of something like this:
```ts
deriveBIP44AddressKey(parentNode, {
  // Both `number` and `{ index: number; isHardened: boolean; }` supported
  change: 0,
  address_index: {
    index: 3333,
    isHardened: true
  }
});
```

FrederikBolding: Yeah, that could also work 👍

rekmarks: I like @Mrtenz's most recent object suggestion, with one tweak: let's rename the `isHardened` property to just `hardened`.

rekmarks: Oh, I'm assuming that if you only specify a number, that will be used as the index. Since this is intended for BIP-44, we should default the hardening behavior to whatever BIP-44 specifies.

src/utils.ts
```diff
@@ -113,6 +116,36 @@ export function getUnhardenedBIP32NodeToken(
   return `bip32:${index}`;
 }
 
+/**
+ * An hardened or unhardened BIP-32 node token, e.g. `bip32:0` or `bip32:0'`.
```
rekmarks, line 120: nit:
```suggestion
 * A hardened or unhardened BIP-32 node token, e.g. `bip32:0` or `bip32:0'`.
```

src/constants.ts
```diff
@@ -11,8 +11,8 @@ export const HEXADECIMAL_KEY_LENGTH = 128 as const;
 export const BASE_64_REGEX =
   /^(?:[A-Za-z0-9+/]{4})*(?:[A-Za-z0-9+/]{3}=|[A-Za-z0-9+/]{2}==)?$/u;
 
-export const MIN_BIP_44_DEPTH = 0 as const;
-export const MAX_BIP_44_DEPTH = 5 as const;
+export const MIN_BIP_44_DEPTH = 0;
+export const MAX_BIP_44_DEPTH = 5;
```
rekmarks, line 15: Why remove the `as const`? Were they redundant?

rekmarks (commented):
LGTM! Just one validation issue and a nit. Submitting this review as a comment so you can resolve them and someone in an EU time zone can approve before merging.

src/constants.ts
Mrtenz: Yeah, TypeScript infers the types of these consts as `0` and `5` respectively without the `as const` assertion too.

src/utils.ts
```diff
@@ -113,6 +116,36 @@ export function getUnhardenedBIP32NodeToken(
   return `bip32:${index}`;
 }
 
+/**
+ * A hardened or unhardened BIP-32 node token, e.g. `bip32:0` or `bip32:0'`.
+ * Validates that the index is a non-negative integer number, and throws an
```
FrederikBolding, line 121: I don't see where it's verified that the index is non-negative? 🤔

src/BIP44CoinTypeNode.test.ts
```diff
@@ -648,6 +759,70 @@ describe('getBIP44AddressKeyDeriver', () => {
...
+        defaultBip39NodeToken,
+        BIP44PurposeNodeToken,
+        `bip32:${coinType}'`,
+        `bip32:0'`,
+        `bip32:2'`,
+        `bip32:0`,
+      ],
+    }).key;
+
+    expect(deriver(0).toString('base64')).toStrictEqual(expectedKey);
```
FrederikBolding, line 823: I wonder if we should use something other than `0` for some of these test values? 🤔 

Just to be on the safe side.

src/utils.ts
```diff
@@ -30,11 +31,13 @@ export function getBIP44CoinTypePathString(
   )}'`;
 }
 
+type BIP44PathIndex = number | { index: number; hardened: boolean };
+
 type BIP44PathIndices = {
   coin_type: number;
```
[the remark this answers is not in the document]
Mrtenz: As per BIP-44, `purpose`, `coin_type` and `account` have to be hardened.

src/BIP44CoinTypeNode.test.ts
Mrtenz: Added a test for this!

src/utils.ts
FrederikBolding: Ah! Can we clarify that?

Mrtenz: Should we clarify implementation details in JSDoc comments?

src/utils.ts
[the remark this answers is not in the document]
FrederikBolding: If we are following BIP-44 to the letter then none of these should be configurable. It's sort of ambiguous in the specification, but let's just keep this for now and we can re-visit if a network comes up that changes the BIP-44 standards even more 😅

src/utils.ts
FrederikBolding: Hmm, actually a good question. I guess not? But it is a bit confusing to have the validation be performed by other functions called 🤔 

It might just be me 😅

FrederikBolding (approved):

PR #38, opened 2022-03-10
Use `@noble/*` libraries

### LINKED ISSUE #36: Use `@noble` packages wherever possible
We should migrate our secp256k1 cryptography to [`@noble/secp256k1`](https://www.npmjs.com/package/@noble/secp256k1).

Mrtenz:
This PR replaces `keccak`, `secp256k1` and the native Node.js `crypto` module with their Noble-equivalents. It also replaces `bip39` with `@scure/bip39` (same author as the Noble libraries).

## To do

- [x] Replace `keccak` with `@noble/hashes`
- [x] Replace Node.js `crypto` module with their Noble-equivalent.

Closes #36.


package.json
```diff
@@ -67,7 +67,8 @@
   "lavamoat": {
     "allowScripts": {
       "@lavamoat/preinstall-always-fail": false,
-      "keccak": true
+      "keccak": true,
+      "secp256k1": true
```
Mrtenz, line 71: @rekmarks `secp256k1` is used by `ethereumjs-wallet` in one of the tests. Couldn't this be replaced by hardcoded test vectors instead? Is there a need to actually call the library in the tests?

test/fixtures.ts
```diff
@@ -314,4 +314,59 @@ export default {
       ],
     },
   ],
+
+  // https://github.com/cryptocoinjs/secp256k1-node/blob/f73cd3a2b95ec6f549a2feabfbf2b8f17b30dcc7/test/privatekey.js
+  // The state of the default branch as of 2022-03-10
+  'secp256k1-node': {
```
FrederikBolding, line 320: Can we write up a bit more how these are extracted?

src/derivers/bip32.ts
```diff
@@ -1,13 +1,13 @@
-import crypto from 'crypto';
```
FrederikBolding, line 1: Potential question for @rekmarks whether we prefer to use `crypto` where possible?

src/derivers/bip32.ts
```diff
@@ -127,6 +114,45 @@ type GenerateKeyArgs = {
   secretExtension: string | Buffer;
 };
 
+/**
+ * Get a BigInt from a byte array.
+ *
+ * @param bytes - The byte array to get the BigInt for.
+ * @returns The byte array as BigInt.
+ */
+function bytesToNumber(bytes: Uint8Array): bigint {
```
FrederikBolding, line 123: Move to `utils.ts`?

Mrtenz: It's not used outside of this file. I can move it if you prefer, though.

FrederikBolding: imo its good practice to place very re-usable code that like in `utils`. But ultimately up to you 😄

src/derivers/bip32.ts
rekmarks: This is a good question; I'm not completely sure. My hunch is that we'd prefer `crypto` when on Node, but it seems like `@scure` / `@noble` has avoided using `crypto`, and I wonder why that is. It seems unlikely that anything else would have better performance.

package.json
rekmarks: Yeah, this is just an accident of history where I first tested against other libraries and later added the actual test vectors. Do you mind creating an issue for this so we can track it and someone can fix it later?

rekmarks (approved):
LGTM. Great work!

package.json
Mrtenz: Will do a PR for this!

PR #40, opened 2022-03-14
Use hardcoded fixtures instead of calling implementations

Mrtenz:
This removes the (dev) dependency on `keccak` and `secp256k1`, and also speeds up the tests a bit. All fixtures are extracted from the implementations.


rekmarks (approved):
Nice work! If you like, you can create a follow-up PR after this to remove the `jest` snapshots wherever we replaced third-party packages with fixtures; those snapshots have now served their purpose.

PR #42, opened 2022-03-16
Replace node symbol with private field

Mrtenz:
It's possible to get all symbols in an object using `Object.getOwnPropertySymbols`, so it's still possible to access the node here. This replaces the symbol with a `#private` field, making it inaccessible.


FrederikBolding (approved):
lgtm - Good catch!

PR #43, opened 2022-03-17
Add support for other curves and implement ed25519

Mrtenz:
This PR adds support for using other curves than `secp256k1`, and implements `ed25519` as one of the new curves.

## Breaking changes

- The constructors for `BIP44Node` and `BIP44CoinTypeNode` are replaced with a static `create` function, which returns a `Promise`.
- All functions used to derive keys now return a `Promise`.


test/reference-implementations.test.ts
```diff
@@ -184,32 +191,136 @@ describe('reference implementation tests', () => {
...
+                );
               }
 
-              expect(targetKey.slice(0, 32).toString('hex')).toStrictEqual(
-                privateKey,
-              );
-            },
+              expect(targetKey.slice(0, 32).toString('hex')).toBe(privateKey);
+              expect(
+                (await ed25519.getPublicKey(targetKey.slice(0, 32))).toString(
```
FrederikBolding, line 248: Is there a reason `getPublicKey` couldn't do the slicing here?

test/fixtures.ts
```diff
@@ -448,4 +482,272 @@ export default {
...
+            publicKey:
+              '0047150c75db263559a70d5778bf36abbab30fb061ad69f69ece61a72b0cfa4fc0',
+          },
+        ],
+      },
+    ],
+
+    // https://github.com/alepop/ed25519-hd-key/tree/d8c0491bc39e197c86816973e80faab54b9cbc26
+    // The state of the default branch as of 2021-10-19
+    'ed25519-hd-key': {
```
FrederikBolding, line 678: Were these generated for testing using this tool? Because I can't find the fixtures referenced in the GitHub repo. If so, can we make that clearer?

src/BIP44Node.ts
```diff
@@ -213,14 +159,20 @@ export class BIP44Node implements BIP44NodeInterface {
...
-  public derive(path: PartialHDPathTuple): BIP44Node {
+  public async derive(path: PartialHDPathTuple): Promise<BIP44Node> {
     if (this.depth === MAX_BIP_44_DEPTH) {
       throw new Error(
         'Illegal operation: This HD tree node is already a leaf node.',
       );
     }
 
-    return deriveChildNode(this.keyBuffer, this.depth, path);
+    const newDepth = this.depth + 1;
```
FrederikBolding, line 169: Are we always guaranteed that the new path is only 1 deeper? 🤔

src/derivation.test.ts
```diff
@@ -63,105 +67,122 @@ describe('derivation', () => {
...
-      parentKey = bip39Derive(mnemonic);
-      parentKey = bip32Derive(`44'`, parentKey);
-      parentKey = bip32Derive(`60'`, parentKey);
-      parentKey = bip32Derive(`0'`, parentKey);
-      parentKey = bip32Derive(`0`, parentKey);
-      const keys = expectedAddresses.map((_, index) => {
-        return bip32Derive(`${index}`, parentKey);
-      });
+
+      /* eslint-disable require-atomic-updates */
```
FrederikBolding, line 173: Why does eslint not like this? 🤔

FrederikBolding: and is it safe to disable 😅

src/BIP44CoinTypeNode.test.ts
```diff
@@ -89,80 +89,99 @@ describe('BIP44CoinTypeNode', () => {
...
         { key: Buffer.allocUnsafe(63).fill(1).toString('base64'), depth: 2 },
         { key: Buffer.alloc(64).toString('base64'), depth: 2 },
-      ].forEach((invalidNode) => {
-        expect(
-          () => new BIP44CoinTypeNode(invalidNode as any, arbitraryCoinType),
-        ).toThrow('Invalid parent key: Must be a non-zero 64-byte key.');
-      });
+      ];
+
+      for (const input of inputs) {
```
FrederikBolding, line 157: Why are we replacing `forEach` with `for .. of ..` here?

src/BIP44Node.test.ts
```diff
@@ -89,333 +89,319 @@ describe('BIP44Node', () => {
...
-          derivationPath: [defaultBip39NodeToken],
-        }).derive([`bip32:43'`]),
-      ).toThrow(
+    it('throws if the depth 1 node of the derivation path is not the BIP-44 purpose node', async () => {
+      const node = await BIP44Node.create({
+        derivationPath: [defaultBip39NodeToken],
+      });
+
+      await expect(() => node.derive([`bip32:43'`])).rejects.toThrow(
         `Invalid derivation path: The "purpose" node node (depth 1) must be the string "${BIP44PurposeNodeToken}".`,
```
FrederikBolding, line 342: This error message reads weird 😅

FrederikBolding: "node node"

src/SLIP10Node.test.ts
```diff
@@ -0,0 +1,345 @@
...
+    it('throws if the key is neither a string nor a buffer', async () => {
+      await expect(() =>
+        SLIP10Node.create({ depth: 0, key: {} as any, curve: secp256k1 }),
+      ).rejects.toThrow(
+        'Invalid key: Must be a Buffer or string if specified. Received: "object"',
+      );
+    });
+
+    it('throws if the key is an invalid Buffer', async () => {
+      const invalidLengthBuffer = Buffer.alloc(63).fill(1);
```
FrederikBolding, line 170: `Buffer.alloc(63, 1)` does the same iirc

FrederikBolding: Not sure if cleaner or not though

src/SLIP10Node.ts
```diff
@@ -0,0 +1,272 @@
...
+  }
+}
+
+/**
+ * Derives a child key from the given parent key.
+ * @param parentKey - The parent key to derive from.
+ * @param parentDepth - The depth of the parent key.
+ * @param pathToChild - The path to the child node / key.
+ * @param curve - The curve to use.
+ * @returns The {@link BIP44Node} corresponding to the derived child key.
```
FrederikBolding, line 249: This doesn't seem right

src/curves/ed25519.ts
```diff
@@ -0,0 +1,17 @@
+import { getPublicKey as getEd25519PublicKey } from '@noble/ed25519';
+
+export { CURVE as curve } from '@noble/ed25519';
+
+export const name = 'ed25519';
+export const secret = Buffer.from('ed25519 seed', 'utf8');
+
+// All private keys are valid for ed25519
```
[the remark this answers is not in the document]
Mrtenz: I don't think I have a good link for this. Ed25519 private keys are just 32 bytes. In the case of Secp256k1 certain private keys can be invalid, but that's not the case for Ed25519.

src/derivers/bip32.ts
```diff
@@ -150,12 +169,18 @@ function generateKey({
...
+  curve,
 }: GenerateKeyArgs) {
   const entropy = hmac(sha512, parentExtraEntropy, secretExtension);
   const keyMaterial = entropy.slice(0, 32);
   // extraEntropy is also called "chaincode"
   const extraEntropy = entropy.slice(32);
-  const privateKey = privateAdd(parentPrivateKey, keyMaterial);
 
+  // If curve is ed25519: The returned child key ki is parse256(IL).
+  if (curve.name === 'ed25519') {
```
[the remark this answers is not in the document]
Mrtenz: This is a special case in SLIP-10, see step 4 here: https://github.com/satoshilabs/slips/blob/master/slip-0010.md#private-parent-key--private-child-key. Does it make sense to generalise this?

test/reference-implementations.test.ts
Mrtenz: `ed25519.getPublicKey` takes a private key, but `targetKey` is a buffer with the private key and chain code appended to it. I don't think it makes sense to handle slicing in the `getPublicKey` function here.

src/derivation.test.ts
Mrtenz:
```
ESLint: Possible race condition: `parentKey` might be reassigned based on an outdated value of `parentKey`.(require-atomic-updates)
```

src/BIP44CoinTypeNode.test.ts
Mrtenz: Can't use `await` with `forEach`.

src/SLIP10Node.test.ts
Mrtenz: We do it like this everywhere, and it's just for the tests, so I think it's fine to keep it like this?

FrederikBolding: Yeah, just food for thought

src/SLIP10Node.test.ts
```diff
@@ -0,0 +1,344 @@
...
+      await expect(() =>
+        SLIP10Node.create({ depth: 1, curve: secp256k1 }),
+      ).rejects.toThrow(
+        'Invalid parameters: Must specify either key or derivation path.',
+      );
+    });
+
+    it('throws if the derivation path is empty', async () => {
+      await expect(() =>
+        SLIP10Node.create({ derivationPath: [] as any, curve: secp256k1 }),
```
ritave, line 143: You could constraint the type using [non-zero arrays](https://stackoverflow.com/a/56006703)

src/curves/ed25519.ts
```diff
@@ -0,0 +1,19 @@
...
+export const isValidPrivateKey = () => true;
```
ritave, line 9: For documentation sake - this should take a single unused parameter, the key

src/derivation.test.ts
ritave: look ok to me, nothing changes `parentKey` in the meantime

src/curves/ed25519.ts
```diff
@@ -0,0 +1,19 @@
+import { getPublicKey as getEd25519PublicKey } from '@noble/ed25519';
+
+export { CURVE as curve } from '@noble/ed25519';
+
+export const name = 'ed25519';
+export const secret = Buffer.from('ed25519 seed', 'utf8');
```
ritave, line 6: I find it weird that the secrets are hard-coded. Shouldn't be they customisable and provided by the extension?

ritave (approved):

src/curves/ed25519.ts
Mrtenz: No, these are [defined by SLIP-10 (and BIP-32)](https://github.com/satoshilabs/slips/blob/master/slip-0010.md#master-key-generation).

src/BIP44Node.ts
```diff
@@ -110,87 +92,51 @@ export class BIP44Node implements BIP44NodeInterface {
...
 
-    let bufferKey: Buffer;
-    if (Buffer.isBuffer(key)) {
-      if (!isValidBufferKey(key)) {
-        throw new Error(
-          'Invalid buffer key: Must be a 64-byte, non-empty Buffer.',
-        );
-      }
+  public get depth(): BIP44Depth {
+    validateBIP44Depth(this.#node.depth);
```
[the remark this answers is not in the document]
Mrtenz: To keep TypeScript happy without the need for a type cast.

src/SLIP10Node.test.ts
Mrtenz: It is constrained, hence the type cast here. Without it it would result in TypeScript errors.

src/BIP44CoinTypeNode.ts
```diff
@@ -106,45 +85,71 @@ export class BIP44CoinTypeNode implements BIP44CoinTypeNodeInterface {
...
+    return this.#node.key;
+  }
+
+  public get keyBuffer(): Buffer {
+    return this.#node.keyBuffer;
+  }
+
+  public readonly path: CoinTypeHDPathString;
+
+  public readonly coin_type: number;
```
rekmarks, line 146: Any particular reason not to keep these at the top of the class declaration?

src/BIP44Node.ts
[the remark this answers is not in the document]
rekmarks: I think it's completely fine to cast here. Even though this validation function is extremely cheap, I think we should avoid using runtime validation in these situations as a matter of course. If we really want to avoid type casts we could genericize the depth in the `SLIP10Node` type, but that's probably best done as a follow-up.

src/BIP44CoinTypeNode.ts
rekmarks: Ah, because static methods can't access instance fields anyway?

src/derivation.test.ts
rekmarks: Yeah, I think it's fine. ESLint is trying to protect us from interweaving async operations and the `parentKey` changing in between our async function calls here. I don't see how that could happen here, so we should be good.

src/BIP44Node.ts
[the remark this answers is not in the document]
rekmarks: I think it's completely fine to cast here, because the validation we already perform convinces us that the internal node is valid.

Furthermore, although this validation function is extremely cheap, I think we should avoid using runtime validation in these situations as a matter of course. If we _really_ want to avoid type casts we could genericize the depth in the `SLIP10Node` type, but that would hardly be worth it, and probably best done as a follow-up.

```suggestion
    this.#node.depth as BIP44Depth;
```

src/SLIP10Node.ts
```diff
@@ -0,0 +1,278 @@
...
+ */
+export async function deriveChildNode(
+  parentKey: Buffer,
+  parentDepth: number,
+  pathToChild: SLIP10PathTuple,
+  curve: Curve,
+): Promise<{ key: Buffer; depth: number }> {
+  if (pathToChild.length === 0) {
+    throw new Error(
+      'Invalid HD tree derivation path: Deriving a path of length 0 is not defined',
```
rekmarks, line 265: Consistency is king:
```suggestion
      'Invalid HD tree derivation path: Deriving a path of length 0 is not defined.',
```

src/SLIP10Node.ts
```diff
@@ -0,0 +1,278 @@
...
+/**
+ * Validates a BIP-32 path depth. Effectively, asserts that the depth is an
+ * integer `number`. Throws an error if validation fails.
+ *
+ * @param depth - The depth to validate.
+ */
+export function validateBIP32Depth(depth: unknown): asserts depth is number {
+  if (typeof depth !== 'number' || !Number.isInteger(depth) || depth < 0) {
+    throw new Error(
+      `Invalid HD tree path depth: The depth must be a positive integer. Received: "${depth}"`,
```
rekmarks, line 244: nit:
```suggestion
      `Invalid HD tree path depth: The depth must be a positive integer. Received: "${depth}".`,
```

src/SLIP10Node.ts
```diff
@@ -0,0 +1,278 @@
...
+      } else if (isValidBase64StringKey(key)) {
+        bufferKey = base64StringToBuffer(key);
+      } else {
+        throw new Error(
+          'Invalid string key: Must be a 64-byte hexadecimal or Base64 string.',
+        );
+      }
+    } else {
+      throw new Error(
+        `Invalid key: Must be a Buffer or string if specified. Received: "${typeof key}"`,
```
rekmarks, line 127: nit:
```suggestion
        `Invalid key: Must be a Buffer or string if specified. Received: "${typeof key}".`,
```

rekmarks (dismissed):
Great work! I think all of my requests for this will be easy enough to address, so feel free to merge this when they're resolved cc: @FrederikBolding @ritave

src/curves/ed25519.ts
Mrtenz: If you directly call `ed25519.isValidPrivateKey` it would, but when doing something like
```ts
const curve: Curve = ed25519;
curve.isValidPrivateKey(key);
```
the parameter just gets ignored. But I will add it anyway just for documentation purposes.

src/curves/ed25519.ts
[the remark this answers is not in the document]
Mrtenz: You're right, didn't know it was mentioned there as well!

src/BIP44CoinTypeNode.ts
Mrtenz: Yeah. Generally I keep static methods before everything else, but if there's another convention we use I'd be happy to change it. Might be something we want to enforce with ESLint too.

ritave (approved):

src/SLIP10Node.ts
```diff
@@ -0,0 +1,289 @@
...
+  if (pathToChild.length === 0) {
+    throw new Error(
+      'Invalid HD tree derivation path: Deriving a path of length 0 is not defined.',
+    );
+  }
+
+  // Note that we do not subtract 1 from the length of the path to the child,
+  // unlike when we calculate the depth of a rooted path.
+  const newDepth = parentDepth + pathToChild.length;
+  validateBIP32Depth(newDepth);
```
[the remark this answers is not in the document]
FrederikBolding: Or derivation path formats?

FrederikBolding (dismissed):
Looks good to me, added one validation question for `SLIP10Node`

FrederikBolding:
@Mrtenz Looks like there is some missing coverage, not sure why the tests aren't showing up in status checks:

https://github.com/MetaMask/key-tree/runs/5676012445?check_suite_focus=true

src/SLIP10Node.ts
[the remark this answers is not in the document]
Mrtenz: SLIP-10 doesn't have any restrictions like BIP-44. It's essentially just BIP-32 with some additional curves.

FrederikBolding (approved):

PR #49, opened 2022-03-24
Add basic SLIP-10 documentation

Mrtenz:
This updates the `README.md` with some basic documentation for `SLIP10Node`.

~Blocked by #43.~

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,7 +94,7 @@ You can derive SLIP-10 keys as follows.
 import { SLIP10Node, secp256k1 } from '@metamask/key-tree';
 
 // Create a SLIP10Node from a derivation path. You can also specify a key and depth instead.
-const node = SLIP10Node.create({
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -84,8 +88,28 @@ const addressKey2 = addressKeyDeriver(2, true);
 // keys and protocol addresses.
 ```
 
+You can derive SLIP-10 keys as follows.
+
+```typescript
+import { SLIP10Node, secp256k1 } from '@metamask/key-tree';
+
+// Create a SLIP10Node from a derivation path. You can also specify a key and depth instead.
+const node = SLIP10Node.create({
```
rekmarks, line 97: Do we know if this is sync or async?

```suggestion
const node = await SLIP10Node.create({
```

FrederikBolding (dismissed):

rekmarks (approved):

PR #50, opened 2022-03-24
Add convenience methods for public keys and addresses

### LINKED ISSUE #45: Add convenience method for getting the public key
For non-Ethereum-like networks we don't have any way to generate addresses, so we should add an easy method to get the public key from a HD node.

Mrtenz:
~~Needs to be rebased after #43 is merged.~~ Done.

This PR adds convenience methods to get the (Ethereum) address and public key from an HD node.

**Why use a function instead of a getter?**
The method for deriving a public key from a private key can be async (in the case of `ed25519` for example), so it had to be an async function.

Closes #45.

### DIFF
--- a/src/SLIP10Node.ts
+++ b/src/SLIP10Node.ts
@@ -154,6 +154,8 @@ export class SLIP10Node implements SLIP10NodeInterface {
   }
 
   get #privateKeyBuffer(): Buffer {
+    // `keyBuffer` consists of both the private key (first 32 bytes) and chain code
+    // (last 32 bytes), so we slice off the chain code here.
     return this.keyBuffer.slice(0, 32);
   }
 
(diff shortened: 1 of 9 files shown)


src/derivers/bip32.ts
```diff
@@ -17,13 +17,13 @@ import { Curve, mod, secp256k1 } from '../curves';
  * @param curve - The curve to use.
```
FrederikBolding, line 17: Delete this part of the doc string 😄

src/SLIP10Node.ts
```diff
@@ -152,6 +153,10 @@ export class SLIP10Node implements SLIP10NodeInterface {
     return bufferToBase64String(this.keyBuffer);
   }
 
+  get #privateKeyBuffer(): Buffer {
+    return this.keyBuffer.slice(0, 32);
```
FrederikBolding, line 157: Can we leave a comment to explain why this is needed at this point? Even if we are refactoring it in a later PR. As discussed on Slack.

FrederikBolding (changes requested):

FrederikBolding (approved):

Issue #46, opened 2022-03-17
Public key derivation


Mrtenz:
Currently we only support `private parent key → private child key` derivation. Deriving public child keys allows for sharing public keys, without needing to expose private keys.

For reference:
- https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki#public-parent-key--public-child-key
- https://github.com/satoshilabs/slips/blob/master/slip-0010.md#public-parent-key--public-child-key

Note that this isn't supported by `ed25519`, as it only works with hardened keys.

rekmarks:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Snaps--Core-615b3a7c08d2b20015eb6c4e/issues/MetaMask/key-tree/46?planning-poker) with ZenHub @FrederikBolding @ritave @hmalik88 @Mrtenz

Issue #44, opened 2022-03-17
Generate extended public and private keys


Mrtenz:
We should add some functionality to get an extended public (`xpub`) and private key (`xprv`). This will require keeping track of the parent fingerprint, index, and depth.

[`@noble/hashes`](https://github.com/paulmillr/noble-hashes) and [`@scure/base`](https://github.com/paulmillr/scure-base) can be used for most of the required cryptographic functions, like RIPEMD-160 and Base58Check.

rekmarks:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Snaps--Core-615b3a7c08d2b20015eb6c4e/issues/MetaMask/key-tree/44?planning-poker) with ZenHub @FrederikBolding @hmalik88 @Mrtenz @ritave

PR #54, opened 2022-04-07
Refactor and separate keys and chain code into separate fields

Mrtenz:
This PR refactors the nodes to split the key into a separate `privateKey` and `chainCode` field. Additionally it also adds a `publicKey` field, which simplifies some of the other code. This is in preparation for adding `xpub`s and `xprv`s (#44), and public key derivation (#46).

Some other things that are changed in this PR:
- The `create` function was replaced with `from[...]` functions.
- The base64 keys are mostly removed, and will be replaced with extended keys in a future PR.
- The `getPublicKey` and `getAddress` functions are replaced with a getter. This is now possible since the public key is stored in the node, rather than being derived from the private key.

To do:
- [x] Add more tests to get the coverage back to 100%.
- [x] Replace `create` function in `BIP44CoinTypeNode`.


src/curves/ed25519.ts
```diff
@@ -15,6 +15,8 @@ export const isValidPrivateKey = (_privateKey: Uint8Array | string | bigint) =>
 
 export const deriveUnhardenedKeys = false;
 
+export const publicKeyLength = 33;
```
FrederikBolding, line 18: These are 1 longer than I expected 🤔

test/fixtures.ts
```diff
@@ -34,6 +34,8 @@ export default {
       '747f302d9c916698912d5f70be53a6cf53bc495803a5523d3a7c3afa2afba94ec3803f838b3e1929ab5481f9da35441372283690fdcf27372c38f40ba134fe03',
     privateKey:
       'f29d6ddd6b0cd1fd59ed99900edd5a53e905b87dfe06824751010feb5228d960',
+    publicKey:
```
FrederikBolding, line 37: Was this just missing before?

test/reference-implementations.test.ts
```diff
@@ -33,29 +36,32 @@ describe('reference implementation tests', () => {
...
             getBIP44CoinTypeToAddressPathTuple({ address_index: index }),
             parentKey,
+            undefined,
+            parentChainCode,
           );
 
+          const publicKey = secp256k1.getPublicKey(childKey, false);
           expect(
-            privateKeyToEthAddress(childKey).toString('hex'),
+            publicKeyToEthAddress(publicKey).toString('hex'),
```
FrederikBolding, line 64: Should we check that we can derive correctly from both priv and pub or is that overkill?

src/constants.ts
```diff
@@ -1,11 +1,13 @@
-export const BUFFER_KEY_LENGTH = 64 as const;
+export const BUFFER_EXTENDED_KEY_LENGTH = 64;
 
-export const BASE_64_KEY_LENGTH = 88 as const;
+export const BUFFER_KEY_LENGTH = 32;
+
+export const BASE_64_KEY_LENGTH = 88;
 
 export const BASE_64_ZERO =
```
FrederikBolding, line 7: If we are gonna remove the base64 parsing (see other question for whether thats intentional), some of these constants can be removed too

src/utils.test.ts
```diff
@@ -0,0 +1,19 @@
+import { isValidBufferKey, nullableHexStringToBuffer } from './utils';
+
+describe('nullableHexStringToBuffer', () => {
+  it('returns undefined when passed undefined', () => {
```
FrederikBolding, line 4: Can we check the other case too?

src/derivers/bip32.ts
```diff
@@ -1,56 +1,83 @@
 import { keccak_256 as keccak256 } from '@noble/hashes/sha3';
 import { hmac } from '@noble/hashes/hmac';
 import { sha512 } from '@noble/hashes/sha512';
-import { BUFFER_KEY_LENGTH, BIP_32_HARDENED_OFFSET } from '../constants';
+import { BIP_32_HARDENED_OFFSET, BUFFER_KEY_LENGTH } from '../constants';
 import { bytesToNumber, hexStringToBuffer, isValidBufferKey } from '../utils';
 import { Curve, mod, secp256k1 } from '../curves';
 
+export function privateKeyToEthAddress(key: Buffer) {
```
FrederikBolding, line 8: Doc string?

src/derivers/bip32.ts
```diff
@@ -1,56 +1,83 @@
...
- * @param key - The `address_index` key buffer to convert to an Ethereum
+ * @param key - The `address_index` public key buffer to convert to an Ethereum
  * address.
  * @returns The Ethereum address corresponding to the given key.
  */
-export function privateKeyToEthAddress(key: Buffer) {
-  if (!Buffer.isBuffer(key) || !isValidBufferKey(key)) {
-    throw new Error('Invalid key: The key must be a 64-byte, non-zero Buffer.');
+export function publicKeyToEthAddress(key: Buffer) {
+  if (!Buffer.isBuffer(key) || !isValidBufferKey(key, 65)) {
```
FrederikBolding, line 29: Could be a constant here too?

src/derivers/bip32.ts
```diff
@@ -1,56 +1,83 @@
...
+    throw new Error('Invalid key: The key must be a 65-byte, non-zero Buffer.');
   }
 
-  const privateKey = key.slice(0, 32);
-  const publicKey = secp256k1.getPublicKey(privateKey, false).slice(1);
-  return Buffer.from(keccak256(Buffer.from(publicKey)).slice(-20));
+  return Buffer.from(keccak256(key.slice(1)).slice(-20));
```
FrederikBolding, line 33: Why do we need to slice the first element here? 🤔

src/derivers/bip32.ts
```diff
@@ -1,56 +1,83 @@
...
  * @param pathPart
  * @param parentKey
+ * @param parentPublicKey
+ * @param chainCode
  * @param curve
  */
 export async function deriveChildKey(
   pathPart: string,
-  parentKey: Buffer,
+  parentKey: Buffer | undefined,
```
FrederikBolding, line 45: `parentKey?: Buffer`?

src/derivers/bip32.ts
```diff
@@ -65,23 +92,25 @@ export async function deriveChildKey(
...
+    parentPrivateKey: parentKey as Buffer,
+    parentExtraEntropy: chainCode,
     secretExtension,
     curve,
   });
 
-  return Buffer.concat([privateKey, extraEntropy]);
+  return [
+    Buffer.from(privateKey),
+    await curve.getPublicKey(privateKey),
```
FrederikBolding, line 111: We should document the change in return value here

src/derivation.ts
```diff
@@ -1,23 +1,28 @@
...
  * ethereum default seed path: "m/44'/60'/0'/0/{account_index}"
  * multipath: "bip32:44'/bip32:60'/bip32:0'/bip32:0/bip32:{account_index}"
  *
  * m: { privateKey, chainCode } = sha512Hmac("Bitcoin seed", masterSeed)
- * 44': { privateKey, chainCode } = parentKey.privateKey + sha512Hmac(parentKey.chainCode, [0x00, parentKey.privateKey, index + HARDENED_OFFSET])
- * 60': { privateKey, chainCode } = parentKey.privateKey + sha512Hmac(parentKey.chainCode, [0x00, parentKey.privateKey, index + HARDENED_OFFSET])
- * 0': { privateKey, chainCode } = parentKey.privateKey + sha512Hmac(parentKey.chainCode, [0x00, parentKey.privateKey, index + HARDENED_OFFSET])
- * 0: { privateKey, chainCode } = parentKey.privateKey + sha512Hmac(parentKey.chainCode, [parentKey.publicKey, index])
- * 0: { privateKey, chainCode } = parentKey.privateKey + sha512Hmac(parentKey.chainCode, [parentKey.publicKey, index])
+ * 44': { privateKey, chainCode } = parentKey.privateKey + sha512Hmac(parentKey.chainCode, [0x00,
```
FrederikBolding, line 16: This formatting change makes it slightly harder to read

src/derivation.test.ts
```diff
@@ -166,24 +174,48 @@ describe('derivation', () => {
...
+        `44'`,
+        privateKey,
+        undefined,
+        chainCode,
+      );
+
+      [privateKey, , chainCode] = await bip32Derive(
+        `60'`,
+        privateKey,
+        undefined,
```
FrederikBolding, line 192: I wonder if we can simplify the overall API so we don't have to have so many `undefined` params, thoughts?

src/SLIP10Node.ts
```diff
@@ -34,142 +37,209 @@ export type JsonSLIP10Node = {
...
-
-type SLIP10NodeConstructorOptions = {
-  readonly depth: number;
-  readonly key: Buffer;
+type SLIP10DerivationPathOptions = {
+  readonly derivationPath: RootedSLIP10PathTuple;
   readonly curve: SupportedCurve;
 };
 
 export class SLIP10Node implements SLIP10NodeInterface {
```
FrederikBolding, line 80: Is it intentional to not be able to create a `SLIP10Node` from a JSON node? Same goes for BIP44

src/BIP44Node.test.ts
```diff
@@ -103,63 +66,17 @@ describe('BIP44Node', () => {
       });
     });
 
-    it('throws if the depth is invalid', async () => {
```
FrederikBolding, line 106: Do we want to remove this test?

src/SLIP10Node.ts
```diff
@@ -34,142 +37,209 @@ export type JsonSLIP10Node = {
...
+  public get chainCode() {
+    return this.chainCodeBuffer.toString('hex');
+  }
+
+  public get privateKey(): string | undefined {
+    return this.privateKeyBuffer?.toString('hex');
+  }
+
+  public get publicKey(): string {
+    return this.publicKeyBuffer.toString('hex');
```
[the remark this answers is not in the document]
Mrtenz: `publicKeyBuffer`, `privateKeyBuffer` and `chainCodeBuffer` are public fields, so they're accessible without a getter. These getters are just a convenient way to get a string representation of the buffers.

src/SLIP10Node.ts
```diff
@@ -34,142 +37,209 @@ export type JsonSLIP10Node = {
...
-      if (!isValidBufferKey(key)) {
-        throw new Error(
-          'Invalid buffer key: Must be a 64-byte, non-empty Buffer.',
-        );
-      }
-
-      bufferKey = key;
-    } else if (typeof key === 'string') {
-      if (isValidHexStringKey(key)) {
-        bufferKey = hexStringToBuffer(key);
```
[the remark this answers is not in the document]
Mrtenz: Base64 strings will be replaced with `xpub`s and `xprv`s, but we could allow hexadecimal strings for the `fromExtendedKey` function?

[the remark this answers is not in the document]
Mrtenz: Base64 strings will be replaced with `xpub`s and `xprv`s, but we could allow hexadecimal strings for the `fromExtendedKey` function?

src/SLIP10Node.ts
```diff
@@ -181,69 +251,45 @@ export class SLIP10Node implements SLIP10NodeInterface {
    * @returns The {@link SLIP10Node} corresponding to the derived child key.
    */
   public async derive(path: SLIP10PathTuple): Promise<SLIP10Node> {
-    const { key, depth } = await deriveChildNode(
-      this.keyBuffer,
+    // TODO: Support deriving from a public key
```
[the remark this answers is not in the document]
Mrtenz: Yes.

src/SLIP10Node.ts
```diff
@@ -181,69 +251,45 @@ export class SLIP10Node implements SLIP10NodeInterface {
...
-
   // This is documented in the interface of this class.
   public toJSON(): JsonSLIP10Node {
     return {
       depth: this.depth,
-      key: this.key,
       curve: this.curve,
+      privateKey: this.privateKey,
+      publicKey: this.publicKey,
+      chainCode: this.chainCodeBuffer.toString('hex'),
```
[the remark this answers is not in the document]
Mrtenz: That's something I added later, and forgot to update this part. Good catch!

src/curves/ed25519.ts
Mrtenz: ED25519 public keys are prefixed with `0x00` for some reason.

test/reference-implementations.test.ts
Mrtenz: The public key is directly derived from the private key, in the same way that the `privateKeyToEthAddress` function would do it, so I don't think it's necessary?

src/derivers/bip32.ts
Mrtenz: The first byte indicates the type of public key (compressed versus uncompressed, and possibly some other factors too). Uncompressed public keys are 65 bytes long for this reason, but we only care about the last 64 bytes.

src/derivers/bip32.ts
Mrtenz: Can't do that here since it's not the last parameter and some parameters after it are required.

src/derivation.test.ts
Mrtenz: We could change it to use options objects, rather than passing everything as a parameter? Or do you have another suggestion for this?

src/BIP44Node.test.ts
Mrtenz: `fromDerivationPath` derives the depth from the derivation path so there's no depth to verify.

src/derivation.test.ts
FrederikBolding: Yeah, that's what I was thinking. But up to you whether you think it improves things or complicates them too much

src/SLIP10Node.ts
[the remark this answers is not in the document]
FrederikBolding: Yeah, I think we'll need that regardless if we want to still support the JSON nodes. Which @rekmarks seemed adamant that we do

test/reference-implementations.test.ts
FrederikBolding: Up to you!

src/derivation.test.ts
Mrtenz: We can definitely do this, but might be better to do it as part of another PR? Since this is already a pretty big refactor, and I want to avoid refactoring even more things.

src/SLIP10Node.ts
```diff
@@ -34,142 +37,209 @@ export type JsonSLIP10Node = {
...
-        throw new Error(
-          'Invalid string key: Must be a 64-byte hexadecimal or Base64 string.',
-        );
-      }
-    } else {
-      throw new Error(
-        `Invalid key: Must be a Buffer or string if specified. Received: "${typeof key}".`,
-      );
+    if (derivationPath) {
+      const [privateKey, , chainCode] = await createKeyFromPath({
```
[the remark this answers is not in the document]
Mrtenz: Yeah, it calls `validateBIP32Depth` to validate the depth and the function called by this calls `validatePathSegment` to validate the segment.

FrederikBolding (approved):
lgtm!

PR #56, opened 2022-04-13
Add support for public key derivation

### LINKED ISSUE #46: Public key derivation

Mrtenz:
Closes #46.

Additional to adding support for public key derivation, I changed the signatures for some functions to use objects, rather than tuples or parameters. This prevents situations where we need to pass `undefined` in place for a private key, for example.

### DIFF
--- a/src/BIP44Node.ts
+++ b/src/BIP44Node.ts
@@ -193,6 +193,20 @@ export class BIP44Node implements BIP44NodeInterface {
   }
 
   /**
+   * Returns a neutered version of this node, i.e. a node without a private key.
+   */
+  public neuter(): BIP44Node {
+    const node = new SLIP10Node({
+      depth: this.depth,
+      chainCode: this.chainCodeBuffer,
+      publicKey: this.publicKeyBuffer,
+      curve: this.#node.curve,
+    });
+
+    return new BIP44Node(node);
+  }
+
+  /**
    * Derives a child of the key contains be this node and returns a new
    * {@link BIP44Node} containing the child key.
    *
--- a/src/BIP44Node.ts
+++ b/src/BIP44Node.ts
@@ -196,13 +196,7 @@ export class BIP44Node implements BIP44NodeInterface {
    * Returns a neutered version of this node, i.e. a node without a private key.
    */
   public neuter(): BIP44Node {
-    const node = new SLIP10Node({
-      depth: this.depth,
-      chainCode: this.chainCodeBuffer,
-      publicKey: this.publicKeyBuffer,
-      curve: this.#node.curve,
-    });
-
+    const node = this.#node.neuter();
     return new BIP44Node(node);
   }
 
--- a/src/curves/curve.ts
+++ b/src/curves/curve.ts
@@ -22,7 +22,7 @@ export type Curve = {
     compressed?: boolean,
   ) => Buffer | Promise<Buffer>;
   isValidPrivateKey: (privateKey: Uint8Array | string | bigint) => boolean;
-  publicAdd: (publicKey: Buffer, tweak: Buffer) => Uint8Array;
+  publicAdd: (publicKey: Buffer, tweak: Buffer) => Buffer;
   compressPublicKey: (publicKey: Buffer) => Buffer;
 };
 
--- a/src/curves/curve.ts
+++ b/src/curves/curve.ts
@@ -22,6 +22,8 @@ export type Curve = {
     compressed?: boolean,
   ) => Buffer | Promise<Buffer>;
   isValidPrivateKey: (privateKey: Uint8Array | string | bigint) => boolean;
+  publicAdd: (publicKey: Buffer, tweak: Buffer) => Uint8Array;
+  compressPublicKey: (publicKey: Buffer) => Buffer;
 };
 
 // As long as both parameters are specified, this function is the same for all curves.
--- a/src/curves/secp256k1.ts
+++ b/src/curves/secp256k1.ts
@@ -25,6 +25,10 @@ export const getPublicKey = (
 
 export const publicAdd = (publicKey: Buffer, tweak: Buffer): Buffer => {
   const point = Point.fromHex(publicKey);
+
+  // The returned child key Ki is point(parse256(IL)) + Kpar.
+  // This multiplies the tweak with the base point of the curve (Gx, Gy).
+  // https://github.com/bitcoin/bips/blob/274fa400d630ba757bec0c03b35ebe2345197108/bip-0032.mediawiki#public-parent-key--public-child-key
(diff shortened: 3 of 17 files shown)


src/derivation.test.ts
```diff
@@ -234,70 +233,62 @@ describe('derivation', () => {
...
-        bip32Derive(`44'`, undefined, undefined, Buffer.alloc(32, 1)),
+        bip32Derive({
+          path: `44`,
+          chainCode: Buffer.alloc(32).fill(1),
+        }),
       ).rejects.toThrow(
         'Invalid parameters: Must specify either a parent private or public key.',
       );
-
-      await expect(
```
FrederikBolding, line 254: Do we not want to test this anymore?

src/derivation.ts
```diff
@@ -43,42 +52,45 @@ import { Curve } from './curves';
...
+export async function deriveKeyFromPath({
+  path,
+  privateKey,
+  publicKey,
+  chainCode,
+  depth,
+  curve,
+}: DeriveKeyFromPathArgs): Promise<DerivedKeys> {
+  if (privateKey && !Buffer.isBuffer(privateKey)) {
     throw new Error('Parent key must be a Buffer if specified.');
```
FrederikBolding, line 64: "Private key" perhaps?

src/derivation.ts
```diff
@@ -94,27 +106,27 @@ function hasDeriver(pathType: string): pathType is keyof typeof derivers {
  * - A lone BIP-39 path node
  * - A multipath
  *
- * @param pathSegment - The path segment string to validate.
+ * @param path - The path segment string to validate.
```
FrederikBolding, line 109: Why are we renaming this?

FrederikBolding: To match the options key?

src/derivers/bip32.ts
```diff
@@ -59,45 +60,35 @@ export function publicKeyToEthAddress(key: Buffer) {
...
     throw new Error(
       `Invalid path: Cannot derive unhardened child keys with ${curve.name}.`,
     );
   }
 
-  if (!parentKey && !parentPublicKey) {
+  if (isHardened && !privateKey) {
     throw new Error(
-      'Invalid parameters: Must specify either a parent private or public key.',
+      'Invalid path: Cannot derive hardened child keys without a private key.',
```
FrederikBolding, line 79: Why change this to "path"? It is "invalid params" isnt it?

src/derivers/bip32.ts
```diff
@@ -111,29 +102,55 @@ export async function deriveChildKey(
...
+  }
+
+  if (publicKey) {
+    if (publicKey.length !== curve.publicKeyLength) {
+      throw new Error(
+        `Invalid parent public key: Must be ${curve.publicKeyLength} bytes long.`,
+      );
+    }
+
+    const compressedPublicKey = curve.compressPublicKey(publicKey);
```
FrederikBolding, line 132: Do we always want to compress the public key?

src/derivers/bip32.ts
```diff
@@ -206,29 +231,64 @@ export function privateAdd(
...
+  publicKey,
+  chainCode,
+  publicExtension,
+  curve,
+}: GeneratePublicKeyArgs): DerivedKeys {
+  const entropy = hmac(sha512, chainCode, publicExtension);
+  const keyMaterial = entropy.slice(0, 32);
+  const childChainCode = entropy.slice(32);
+
+  // This function may fail if the resulting key is invalid.
```
FrederikBolding, line 287: Does it throw then? If so, can we test that?

src/curves/curve.ts
```diff
@@ -22,6 +22,8 @@ export type Curve = {
     compressed?: boolean,
   ) => Buffer | Promise<Buffer>;
   isValidPrivateKey: (privateKey: Uint8Array | string | bigint) => boolean;
+  publicAdd: (publicKey: Buffer, tweak: Buffer) => Uint8Array;
```
FrederikBolding, line 25: Any reason for this not to return a `Buffer`?

src/curves/secp256k1.ts
```diff
@@ -18,3 +23,20 @@ export const getPublicKey = (
   privateKey: Uint8Array | string | bigint,
   compressed?: boolean,
 ): Buffer => Buffer.from(getSecp256k1PublicKey(privateKey, compressed));
+
+export const publicAdd = (publicKey: Buffer, tweak: Buffer): Uint8Array => {
+  const point = Point.fromHex(publicKey);
+  const newPoint = point.add(
+    // Multiplies `Point(Gx, Gy)` with the tweak
```
FrederikBolding, line 30: Can we link a source on this?

src/derivation.ts
Mrtenz: Before we would have `pathSegment`, `pathToChild`, `pathValue`, `pathPart`, etc. depending on which function you are using. I changed it all to `path` for consistency.

src/BIP44Node.ts
```diff
@@ -192,6 +192,20 @@ export class BIP44Node implements BIP44NodeInterface {
     Object.freeze(this);
   }
 
+  /**
+   * Returns a neutered version of this node, i.e. a node without a private key.
+   */
+  public neuter(): BIP44Node {
```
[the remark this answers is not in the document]
Mrtenz: Yes 😅 .

src/curves/secp256k1.ts
Mrtenz: Simplified this a bit. Do you still want me to link something?

src/derivation.test.ts
Mrtenz: It's tested in a separate `it` block below, so I removed the duplicate test here.

src/derivers/bip32.ts
FrederikBolding: Would be cool to test if possible, but otherwise fine.

src/derivers/bip32.ts
Mrtenz: Yeah, the child public key is derived from the compressed parent public key.
- https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki#conventions
- https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki#public-parent-key--public-child-key

FrederikBolding (approved):

PR #58, opened 2022-04-21
Add extended public (`xpub`) and private (`xprv`) keys

### LINKED ISSUE #44: Generate extended public and private keys

Mrtenz:
Closes #44.

This PR adds a getter to the `BIP44Node` class to get an extended public or private key, and expands the `fromExtendedKey` function to make it possible to create a node from a stringified extended public or private key.

To make this possible, two new fields are introduced in `SLIP10Node`, `index` and `parentFingerprint`. These are part of the extended keys, so need to be kept track of. Additionally, to reduce complexity in the derivation code, all derivation functions now return a `SLIP10Node` instance, rather than an object containing the new key, chain code, etc.

Note that extended keys are not available for `SLIP10Node`s, since they are not defined in SLIP-10. Its only possible to use extended keys with BIP-44 nodes using `secp256k1`.

### DIFF
--- a/src/BIP44CoinTypeNode.ts
+++ b/src/BIP44CoinTypeNode.ts
@@ -77,6 +77,8 @@ export class BIP44CoinTypeNode implements BIP44CoinTypeNodeInterface {
 
     const node = await BIP44Node.fromExtendedKey({
       depth: json.depth,
+      index: json.index,
+      parentFingerprint: json.parentFingerprint,
       chainCode: hexStringToBuffer(json.chainCode),
       privateKey: nullableHexStringToBuffer(json.privateKey),
       publicKey: hexStringToBuffer(json.publicKey),
@@ -190,6 +192,14 @@ export class BIP44CoinTypeNode implements BIP44CoinTypeNodeInterface {
 
   public get address(): string {
     return this.#node.address;
+  }
+
+  public get parentFingerprint(): number {
+    return this.#node.parentFingerprint;
+  }
+
+  public get index(): number {
+    return this.#node.index;
   }
 
   /**
--- a/src/BIP44CoinTypeNode.ts
+++ b/src/BIP44CoinTypeNode.ts
@@ -186,6 +186,10 @@ export class BIP44CoinTypeNode implements BIP44CoinTypeNodeInterface {
     return this.#node.publicKey;
   }
 
+  public get compressedPublicKeyBuffer(): Buffer {
+    return this.#node.compressedPublicKeyBuffer;
+  }
+
   public get chainCode(): string {
     return this.#node.chainCode;
   }
@@ -196,6 +200,10 @@ export class BIP44CoinTypeNode implements BIP44CoinTypeNodeInterface {
 
   public get parentFingerprint(): number {
     return this.#node.parentFingerprint;
+  }
+
+  public get fingerprint(): number {
+    return this.#node.fingerprint;
   }
 
   public get index(): number {
--- a/src/BIP44CoinTypeNode.ts
+++ b/src/BIP44CoinTypeNode.ts
@@ -17,8 +17,8 @@ import {
   hexStringToBuffer,
   nullableHexStringToBuffer,
 } from './utils';
-import { secp256k1 } from './curves';
-import { ChildNode, deriveChildNode } from './SLIP10Node';
+import { deriveChildNode, SLIP10Node } from './SLIP10Node';
+import { SupportedCurve } from './curves';
 
 export type CoinTypeHDPathTuple = [
   BIP39Node,
@@ -202,6 +202,10 @@ export class BIP44CoinTypeNode implements BIP44CoinTypeNodeInterface {
     return this.#node.index;
   }
 
+  public get curve(): SupportedCurve {
+    return this.#node.curve;
+  }
+
   /**
    * Derives a BIP-44 `address_index` key corresponding to the path of this
    * node and the specified `account`, `change`, and `address_index` values.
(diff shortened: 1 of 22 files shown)


src/BIP44CoinTypeNode.ts
```diff
@@ -291,33 +306,28 @@ export async function deriveBIP44AddressKey(
   // TODO: Support xpubs and xprvs as part of a separate PR.
```
FrederikBolding, line 306: This should be done now 😄

src/SLIP10Node.ts
```diff
@@ -119,12 +139,15 @@ export class SLIP10Node implements SLIP10NodeInterface {
 
     validateCurve(curve);
     validateBIP32Depth(depth);
+    validateParentFingerprint(parentFingerprint);
```
FrederikBolding, line 142: Do we need to validate `index`?

src/curves/secp256k1.ts
```diff
@@ -40,3 +40,8 @@ export const compressPublicKey = (publicKey: Uint8Array): Buffer => {
   const point = Point.fromHex(publicKey);
   return Buffer.from(point.toRawBytes(true));
 };
+
+export const decompressPublicKey = (publicKey: Uint8Array): Buffer => {
+  const point = Point.fromHex(publicKey);
```
FrederikBolding, line 45: Does `Point.fromHex` do the decompression here or where is it happening? Perhaps leave a comment to explain.

src/derivation.test.ts
```diff
@@ -235,61 +249,36 @@ describe('derivation', () => {
...
-          path: `44'`,
-          privateKey: Buffer.alloc(32, 1),
-        }),
-      ).rejects.toThrow('Invalid parameters: Must specify a chain code.');
-    });
+    it('throws when trying to derive from a public key node', async () => {
+      const node = await bip39Derive({ path: mnemonic });
+      const publicNode = node.neuter();
 
-    it('throws for an invalid chain code', async () => {
```
FrederikBolding, line 285: How come we don't test the chain code validation anymore?

src/derivation.test.ts
```diff
@@ -235,61 +249,36 @@ describe('derivation', () => {
...
-          chainCode: Buffer.alloc(32).fill(1),
+          path: '0',
         }),
       ).rejects.toThrow(
-        'Invalid parameters: Must specify either a parent private or public key.',
+        'Invalid parameters: Must specify a node to derive from.',
       );
     });
 
-    it('throws for an invalid private key', async () => {
```
FrederikBolding, line 256: Why are we removing these tests?

src/utils.ts
```diff
@@ -292,3 +305,36 @@ function validateBuffer(
...
+    throw new Error(
+      `Invalid value: Value is not base58-encoded, or the checksum is invalid.`,
+    );
+  }
+};
+
+export const encodeBase58check = (value: Buffer): string => {
+  const base58Check = scureBase58check(sha256);
+
+  return base58Check.encode(value);
```
FrederikBolding, line 324: Can this throw? If so, should we wrap it?

src/extended-keys.ts
```diff
@@ -0,0 +1,146 @@
+import {
+  decodeBase58check,
+  encodeBase58check,
+  isValidBufferKey,
+} from './utils';
+import { validateBIP44Depth } from './BIP44Node';
+import { compressPublicKey, decompressPublicKey } from './curves/secp256k1';
+
+export const PUBLIC_KEY_VERSION = 0x0488b21e;
```
FrederikBolding, line 9: Link to a reference for these please!

src/derivers/bip39.ts
```diff
@@ -35,13 +33,17 @@ export async function deriveChildKey({
...
+  const chainCode = key.slice(32);
 
-  return {
+  return SLIP10Node.fromExtendedKey({
     privateKey,
-    publicKey: await curve.getPublicKey(privateKey),
-    chainCode: key.slice(32),
-  };
+    chainCode,
+    depth: 0,
```
FrederikBolding, line 44: Is the `depth` supposed to be 0 here? 🤔

src/derivers/bip32.ts
```diff
@@ -52,42 +53,33 @@ export function publicKeyToEthAddress(key: Buffer) {
  * Derive a BIP-32 child key with a given path from a parent key.
  *
  * @param pathPart - The derivation path part to derive.
```
FrederikBolding, line 55: Update the doc string

src/derivers/bip32.ts
```diff
@@ -249,7 +242,7 @@ async function generateKey({
   chainCode,
   secretExtension,
   curve,
-}: GenerateKeyArgs): Promise<DerivedKeys> {
+}: GenerateKeyArgs): Promise<DerivedKeys & { privateKey: Buffer }> {
```
FrederikBolding, line 245: Does this overwrite the current `privateKey` type completely?

src/derivation.ts
```diff
@@ -41,38 +52,42 @@ type DeriveKeyFromPathArgs = {
...
-    throw new Error('Private key must be a Buffer if specified.');
+export async function deriveKeyFromPath(
+  args: DeriveKeyFromPathArgs,
+): Promise<SLIP10Node> {
+  const { path, depth = path.length } = args;
+
+  const node = 'node' in args ? args.node : undefined;
+  const curve = 'curve' in args ? args.curve : node?.curve;
+
+  if (!curve) {
```
FrederikBolding, line 72: Do we not need to check the validity of `node` somewhere here? We had checks for the private key validity before.

src/SLIP10Node.ts
Mrtenz: Yes, we should

src/utils.ts
Mrtenz: AFAIK this just accepts any arbitrary Buffer, so it should never throw.

src/derivers/bip39.ts
Mrtenz: Yes, the master node has a depth of 0.

src/derivation.test.ts
Mrtenz: The function doesn't accept a plain chain code anymore, and `SLIP10Node` validates any inputs upon creating the node.

src/derivation.test.ts
Mrtenz: The function doesn't accept a plain private key anymore, and `SLIP10Node` validates any inputs upon creating the node.

FrederikBolding (approved):

PR #69, opened 2022-09-05
Allow a depth larger than 5 for SLIP-10 nodes

Mrtenz:
SLIP-10 nodes can have any depth. There was some left over code from BIP-44, which checks if the depth is at most the maximum BIP-44 depth, but this is not accurate for SLIP-10. `BIP44Node` has its own checks for the max node depth, so it should be fine removing this.


src/derivation.ts
```diff
@@ -141,12 +141,6 @@ export function validatePathSegment(
     throw new Error(`Invalid HD path segment: The segment must not be empty.`);
   }
 
-  if (path.length - 1 > MAX_BIP_44_DEPTH) {
```
FrederikBolding, line 144: Is this validated elsewhere for BIP44?

FrederikBolding (commented):
Looks like CI is failing as well.

src/derivation.ts
Mrtenz: > `BIP44Node` has its own checks for the max node depth, so it should be fine removing this.

FrederikBolding (approved):

PR #74, opened 2022-09-16
Replace postinstall script with Yarn plugin

### LINKED ISSUE #73: Package installation fails because of post install script
Upgrading to the latest version 5.0.0 fails with this error:

```
command not found: simple-git-hooks
```

Probably because of this commit: c1b9bdd6ec254e44bb6a0ef41d534884a71cd1b4 which adds a post install script triggering `simple-git-hooks`. Since you migrated to Yarn 3 for this package I suggest taking a look at how Husky's doc advises the configuration of such a post install script for newer versions of Yarn: https://typicode.github.io/husky/#/?id=yarn-2

Mrtenz:
Closes #73.

This replaces the `postinstall` script with a Yarn plugin which calls `simple-git-hooks` after running `yarn`.

### DIFF
--- a/.yarn/plugins/plugin-git-hooks.js
+++ b/.yarn/plugins/plugin-git-hooks.js
@@ -1,14 +1,7 @@
-const { execute } = require('@yarnpkg/shell');
+const { execSync } = require('child_process');
 
-const afterAllInstalled = async () => {
-  const exitCode = await execute('yarn simple-git-hooks');
-
-  if (exitCode !== 0) {
-    // We have to use `process.exit` here rather than setting `process.exitCode`
-    // because Yarn will override any exit code set in this hook.
-    // eslint-disable-next-line node/no-process-exit
-    process.exit(exitCode);
-  }
+const afterAllInstalled = (project) => {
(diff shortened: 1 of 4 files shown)


.yarn/plugins/plugin-git-hooks.js
```diff
@@ -0,0 +1,18 @@
+const { execSync } = require('child_process');
+
+const afterAllInstalled = (project) => {
+  execSync('yarn simple-git-hooks', { cwd: project.cwd, stdio: 'inherit' });
```
FrederikBolding, line 4: Shouldn't we do this the same way we do it here? https://github.com/LavaMoat/LavaMoat/blob/main/packages/yarn-plugin-allow-scripts/sources/index.ts#L7

Mrtenz: This doesn't seem to work for local plugins:
```
Run yarn --immutable
Internal Error: Cannot find module '@yarnpkg/shell'
Require stack:
- /home/runner/work/key-tree/key-tree/.yarn/plugins/plugin-git-hooks.js
```

FrederikBolding (approved):

PR #77, opened 2022-09-19
fix finger print

amalia20220301:


Mrtenz:
Thanks for fixing the fingerprint logic! I suppose we should add some reference tests for this, to make sure that the current implementation is correct.

Mrtenz (approved):

FrederikBolding (changes requested):
Let's add some reference tests to this before merging, so we are sure we do it correctly now 😅

Mrtenz:
Superseded by #78.

PR #83, opened 2022-10-12
Swap out Buffer with browser-compatible Uint8Array

Mrtenz:
`Buffer` is a Node.js API and does not work in browsers without polyfills. This swaps out all usage of `Buffer` with `Uint8Array`, to make this library fully browser compatible without the need for polyfills.

### DIFF
--- a/src/SLIP10Node.ts
+++ b/src/SLIP10Node.ts
@@ -1,3 +1,4 @@
+import { bytesToHex } from '@metamask/utils';
 import {
   BUFFER_KEY_LENGTH,
   RootedSLIP10PathTuple,
@@ -64,18 +65,18 @@ export type JsonSLIP10Node = {
 };
 
 export type SLIP10NodeInterface = JsonSLIP10Node & {
-  chainCodeBuffer: Buffer;
+  chainCodeBuffer: Uint8Array;
 
   /**
    * The private key for this node, as a Node.js Buffer or browser-equivalent.
    * May be undefined if this node is a public key only node.
    */
-  privateKeyBuffer?: Buffer;
+  privateKeyBuffer?: Uint8Array;
 
   /**
    * The public key for this node, as a Node.js Buffer or browser-equivalent.
    */
-  publicKeyBuffer: Buffer;
+  publicKeyBuffer: Uint8Array;
 
   /**
    * @returns A JSON-compatible representation of this node's data fields.
@@ -88,9 +89,9 @@ type SLIP10NodeConstructorOptions = {
   readonly masterFingerprint?: number;
   readonly parentFingerprint: number;
   readonly index: number;
-  readonly chainCode: Buffer;
-  readonly privateKey?: Buffer;
-  readonly publicKey: Buffer;
+  readonly chainCode: Uint8Array;
+  readonly privateKey?: Uint8Array;
+  readonly publicKey: Uint8Array;
   readonly curve: SupportedCurve;
 };
 
@@ -99,9 +100,9 @@ type SLIP10ExtendedKeyOptions = {
   readonly masterFingerprint?: number;
   readonly parentFingerprint: number;
   readonly index: number;
-  readonly chainCode: string | Buffer;
-  readonly privateKey?: string | Buffer;
-  readonly publicKey?: string | Buffer;
+  readonly chainCode: string | Uint8Array;
+  readonly privateKey?: string | Uint8Array;
+  readonly publicKey?: string | Uint8Array;
   readonly curve: SupportedCurve;
 };
 
@@ -130,6 +131,9 @@ export class SLIP10Node implements SLIP10NodeInterface {
    * validation fails.
    *
    * @param depth - The depth of the node.
+   * @param masterFingerprint - The fingerprint of the master node, i.e., the
+   * node at depth 0. May be undefined if this node was created from an extended
+   * key.
    * @param parentFingerprint - The fingerprint of the parent key, or 0 if
    * the node is a master node.
    * @param index - The index of the node, or 0 if the node is a master node.
@@ -166,7 +170,7 @@ export class SLIP10Node implements SLIP10NodeInterface {
         index,
         chainCode: chainCodeBuffer,
         privateKey: privateKeyBuffer,
-        publicKey: await getCurveByName(curve).getPublicKey(privateKey),
+        publicKey: await getCurveByName(curve).getPublicKey(privateKeyBuffer),
         curve,
       });
     }
@@ -247,11 +251,11 @@ export class SLIP10Node implements SLIP10NodeInterface {
 
   public readonly index: number;
 
-  public readonly chainCodeBuffer: Buffer;
+  public readonly chainCodeBuffer: Uint8Array;
 
-  public readonly privateKeyBuffer?: Buffer;
+  public readonly privateKeyBuffer?: Uint8Array;
 
-  public readonly publicKeyBuffer: Buffer;
+  public readonly publicKeyBuffer: Uint8Array;
 
   constructor({
     depth,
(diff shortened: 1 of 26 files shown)


src/SLIP10Node.ts
```diff
@@ -276,23 +280,27 @@ export class SLIP10Node implements SLIP10NodeInterface {
...
+    if (this.privateKeyBuffer) {
+      return bytesToHex(this.privateKeyBuffer);
+    }
+
+    return undefined;
   }
 
   public get publicKey(): string {
-    return this.publicKeyBuffer.toString('hex');
+    return bytesToHex(this.publicKeyBuffer);
```
FrederikBolding, line 295: All of these will be 0x prefixed now, if that is intended we should make sure to let Snap developers know when we ship this.

src/curves/curve.ts
```diff
@@ -18,13 +18,13 @@ export type Curve = {
     n: bigint;
   };
   getPublicKey: (
-    privateKey: Uint8Array | string | bigint,
```
FrederikBolding, line 21: Why are we limiting support here? 🤔

src/derivers/bip32.ts
```diff
@@ -166,22 +172,20 @@ async function deriveSecretExtension({
...
+    view.setUint32(0, childIndex + BIP_32_HARDENED_OFFSET, false);
+    return concatBytes([new Uint8Array([0]), privateKey, indexBytes]);
   }
 
   // Normal child
-  const indexBuffer = Buffer.allocUnsafe(4);
-  indexBuffer.writeUInt32BE(childIndex, 0);
   const parentPublicKey = await curve.getPublicKey(privateKey, true);
-  return Buffer.concat([parentPublicKey, indexBuffer]);
+  return derivePublicExtension({ parentPublicKey, childIndex, curve });
```
FrederikBolding, line 184: Why is the curve included here?

src/derivers/bip32.ts
```diff
@@ -276,14 +284,14 @@ function generatePublicKey({
   curve,
 }: GeneratePublicKeyArgs): DerivedKeys {
   const entropy = hmac(sha512, chainCode, publicExtension);
-  const keyMaterial = entropy.slice(0, 32);
-  const childChainCode = entropy.slice(32);
+  const keyMaterial = entropy.subarray(0, 32);
```
FrederikBolding, line 287: Are subarrays safe for this?

src/utils.test.ts
```diff
@@ -208,66 +206,39 @@ describe('isHardened', () => {
...
-    expect(hexStringToBuffer('0x1234')).toStrictEqual(
-      Buffer.from('1234', 'hex'),
-    );
+    expect(hexStringToBuffer('1234')).toStrictEqual(hexToBytes('1234'));
+    expect(hexStringToBuffer('0x1234')).toStrictEqual(hexToBytes('1234'));
   });
 
   it('throws if the string is not a valid hex string', () => {
-    expect(() => hexStringToBuffer('')).toThrow('Invalid hex string: "".');
-    expect(() => hexStringToBuffer('0x')).toThrow('Invalid hex string: "0x".');
```
FrederikBolding, line 251: Do we support `0x` now?

src/SLIP10Node.ts
Mrtenz: That's intentional, and yeah, we need to communicate that.

I think in general we should standardise hexadecimal values everywhere to start with `0x`. The problem is otherwise that we have to guess what the user meant. Values like `0x1234` are clearly hexadecimal, but `1234` could be interpreted as a number as well. The same goes for UTF-8-like values.

src/curves/curve.ts
Mrtenz: It looks like the underlying implementation doesn't support `string` or `bigint`? I was running into errors, even though the type says it's supported. Regardless, it's better to just accept `Uint8Arrays`, as otherwise we run into the same problem as mentioned above for string values.

src/derivers/bip32.ts
Mrtenz: It's part of the type. Might have been used at some point, but I can remove it now, I guess.

src/utils.test.ts
Mrtenz: Yes. `hexToBytes('0x')` is equivalent to `new Uint8Array()`.

src/SLIP10Node.ts
```diff
@@ -247,11 +251,11 @@ export class SLIP10Node implements SLIP10NodeInterface {
...
   public readonly index: number;
 
-  public readonly chainCodeBuffer: Buffer;
+  public readonly chainCodeBuffer: Uint8Array;
 
-  public readonly privateKeyBuffer?: Buffer;
+  public readonly privateKeyBuffer?: Uint8Array;
 
-  public readonly publicKeyBuffer: Buffer;
+  public readonly publicKeyBuffer: Uint8Array;
```
rekmarks, line 258: This change will be breaking. Should we rename these variables to `*Bytes` instead of `*Buffer`?

src/SLIP10Node.ts
rekmarks: I second this. If it ain't `0x`-prefixed it better be a local variable with a descriptive name.

src/extended-keys.ts
```diff
@@ -55,29 +55,37 @@ export const decodeExtendedKey = (extendedKey: string): ExtendedKey => {
...
   }
 
+  const keyView = new DataView(key.buffer, key.byteOffset, key.byteLength);
+
   if (version === PUBLIC_KEY_VERSION) {
-    if (key.readUInt8(0) !== 0x02 && key.readUInt8(0) !== 0x03) {
+    if (keyView.getUint8(0) !== 0x02 && keyView.getUint8(0) !== 0x03) {
       throw new Error(
         `Invalid extended key: Public key must start with 0x02 or 0x03.`,
       );
```
rekmarks, line 91: Why do we have a leading zero in the hex numbers here and below?

rekmarks (changes requested):
In addition to my inline comment(s), there are some questions by @FrederikBolding that should be resolved before we merge this.

src/derivers/bip32.ts
rekmarks: Are you going to remove this?

src/utils.ts
```diff
@@ -172,49 +172,25 @@ export function isHardened(bip32Token: string): boolean {
...
- * may or may not be `0x`-prefixed, and the test is case-insensitive.
- *
- * @param hexString - The string to test.
- * @returns Whether the specified string is a valid hexadecimal string. The
- * string may or may not be `0x`-prefixed.
- */
-export function isValidHexString(hexString: string): boolean {
-  return /^(?:0x)?[a-f0-9]+$/iu.test(hexString);
-}
-
```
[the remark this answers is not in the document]
Mrtenz: `@metmask/utils` has the same functionality (`isValidHex` and `hexToBytes`), so there is no need to keep these around in `key-tree`.

src/extended-keys.ts
Mrtenz: While it would technically be the same, it's common to pad a hexadecimal string with an extra `0` when working with bytes. A single hexadecimal can be used to represent a nibble (4 bits), while two hexadecimal characters can be used to represent a byte. JavaScript interprets `0x02` and `0x03` as `2` and `3` regardless of the number of leading zeroes.

src/derivers/bip32.ts
Mrtenz: Done!

src/derivers/bip32.ts
Mrtenz: It's not used anymore after it's returned from this function, so in theory it should be fine. If we want to be on the safe side, we could return a copy of the original byte array in the actual `SLIP10Node` class, but that might be overkill?

rekmarks: Nah, let's create fresh copies in the `SLIP10Node` class. Seems like a bad idea to have byte data with lifecycles persisting across multiple functions/contexts.

rekmarks: Alternatively, since we're working with pretty tiny amounts of memory, we could just use `.slice()` everywhere. I don't know that we'll see a significant performance difference either way.

src/SLIP10Node.ts
rekmarks: We're gonna do this in a follow-up.

rekmarks (approved):
LGTM!

PR #84, opened 2022-10-24
BREAKING: Rename all occurrences of `Buffer` with `Bytes` or `Uint8Array`

Mrtenz:
To make it clear that we no longer use Node.js `Buffer`s, I renamed all usage of `Buffer` to `Bytes` or `Uint8Array` (depending on context).


rekmarks (approved):
LGTM!

PR #94, opened 2022-12-05
Bump ESLint configs and related dependencies

Mrtenz:
This bumps ESLint and configs to the latest version, and fixes linting issues.


.eslintrc.js
```diff
@@ -69,5 +103,5 @@ module.exports = {
     },
   ],
 
-  ignorePatterns: ['!.eslintrc.js', '!.prettierrc.js', 'dist/'],
+  ignorePatterns: ['!.eslintrc.js', '!.prettierrc.js', 'dist/', 'docs/'],
```
mcmire, line 106: Good call, maybe we should do this for the module template.

src/BIP44CoinTypeNode.ts
```diff
@@ -147,7 +147,8 @@ export class BIP44CoinTypeNode implements BIP44CoinTypeNodeInterface {
     validateCoinType(coin_type);
     validateCoinTypeNodeDepth(node.depth);
 
-    return new BIP44CoinTypeNode(node, coin_type);
+    // TODO: Make this function not async in a future version.
+    return Promise.resolve(new BIP44CoinTypeNode(node, coin_type));
```
mcmire, line 151: Why was this failing lint? This method should already return a promise, no?

src/derivation.ts
```diff
@@ -169,6 +174,11 @@ export function validatePathSegment(
...
 }
 
+/**
+ * Get the error for a malformed path segment.
+ *
+ * @returns The error.
+ */
 function getMalformedError() {
-  throw new Error('Invalid HD path segment: The path segment is malformed.');
+  return new Error('Invalid HD path segment: The path segment is malformed.');
```
mcmire, line 183: Oooh, good catch!

tsconfig.json
```diff
@@ -9,5 +9,5 @@
     "strict": true,
     "target": "ES2017"
   },
-  "include": ["src"]
+  "include": ["src", "test"]
```
mcmire, line 12: With this setting, could test files end up in the build without us knowing?

What do you think about defining two tsconfigs, one for dev and one for building? See: https://github.com/MetaMask/metamask-module-template/blob/main/tsconfig.build.json#L11

mcmire (commented):
The latest updates to the ESLint rules are great, looks like we can clean up a lot of stuff. Just had one thing, really.

src/BIP44CoinTypeNode.ts
Mrtenz: `async` functions must use `await` or return a `Promise` in the function. This function previously didn't have any async logic, and it should be changed to not return a `Promise`, but doing so would be a breaking change.

tsconfig.json
Mrtenz: Done!

.eslintrc.js
Mrtenz: Looks like this is already done in the module template.

mcmire: Ah my bad. Cool :)

PR #101, opened 2023-01-04
Update `bip39` implementation to `@metamask/scure-bip39`

adonesky1:
We are updating all instances of `bip39` across our packages to use the `@metamask/scure-bip39` (our fork of `@scure/bip39`) - which allows us to pass around SRP's as `Uint8Array`s rather than plain strings for security purposes.

Currently this change may also be blocking: https://github.com/MetaMask/metamask-extension/pull/17056

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     ]
   },
   "dependencies": {
-    "@metamask/scure-bip39": "^2.0.4",
+    "@metamask/scure-bip39": "^2.1.0",
     "@metamask/utils": "^3.3.0",
     "@noble/ed25519": "^1.6.0",
     "@noble/hashes": "^1.0.0",
--- a/package.json
+++ b/package.json
@@ -41,8 +41,7 @@
     "@noble/ed25519": "^1.6.0",
     "@noble/hashes": "^1.0.0",
     "@noble/secp256k1": "^1.5.5",
-    "@scure/base": "^1.0.0",
-    "@scure/bip39": "^1.0.0"
+    "@scure/base": "^1.0.0"
   },
   "devDependencies": {
     "@lavamoat/allow-scripts": "^2.0.3",
--- a/package.json
+++ b/package.json
@@ -36,6 +36,7 @@
     ]
   },
   "dependencies": {
+    "@metamask/scure-bip39": "^2.0.4",
     "@metamask/utils": "^3.3.0",
     "@noble/ed25519": "^1.6.0",
     "@noble/hashes": "^1.0.0",
--- a/src/derivers/bip39.ts
+++ b/src/derivers/bip39.ts
@@ -1,4 +1,4 @@
-import { mnemonicToSeedSync } from '@metamask/scure-bip39';
+import { mnemonicToSeed } from '@metamask/scure-bip39';
 import { wordlist as englishWordlist } from '@metamask/scure-bip39/dist/wordlists/english';
 import { hmac } from '@noble/hashes/hmac';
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -36,6 +36,7 @@
     ]
   },
   "dependencies": {
+    "@metamask/scure-bip39": "^2.0.4",
```
FrederikBolding, line 39: Shouldn't we remove `@scure/bip39` then too? 😅

FrederikBolding (commented):
The internals of this library will still expect the SRP to be passed as a string. We will need to make further changes to eliminate that assumption.

cc @Mrtenz

package.json
adonesky1: done [here](https://github.com/MetaMask/key-tree/pull/101/commits/24f59fae8a237638242988b3339a6e3e6b22b092)

src/derivers/bip39.ts
```diff
@@ -30,7 +31,10 @@ export async function deriveChildKey({
   path,
   curve,
 }: DeriveChildKeyArgs): Promise<SLIP10Node> {
-  return createBip39KeyFromSeed(await mnemonicToSeed(path), curve);
+  return createBip39KeyFromSeed(
+    mnemonicToSeedSync(path, englishWordlist),
```
Gudahtt, line 35: Hmm, why the switch to the synchronous version of this function?

adonesky1: We [did not update the async version](https://github.com/MetaMask/scure-bip39/blob/main/src/index.ts#L137) of the function, I don't remember why, but I assumed we wanted to start using the sync version given this fact.

adonesky1: Perhaps we should update the async version too?

Gudahtt: Ah, good catch.

This might be a genuine performance concern. The async implementation will use `nextTick` to split the work up to avoid blocking the main thread. Key generation can be slow on some machines.

I guess this works for now, but it might be worth our time to update the async version of this function as well.

Gudahtt (dismissed):
LGTM!

Gudahtt:
Could you elaborate on why this might be blocking that extension PR? I'm having trouble finding a connection.

adonesky1:
Not going to merge this yet @Gudahtt, @FrederikBolding has indicated that there may be some other adaptations required to make sure this is not breaking. I'm meeting with @Mrtenz tomorrow to discuss.

adonesky1:
> Could you elaborate on why this might be blocking that extension PR? I'm having trouble finding a connection.

I'm not sure if this is going to unblock that but it may be related. The error thrown [here](https://app.circleci.com/pipelines/github/MetaMask/metamask-extension/33064/workflows/ce72434a-bbc5-4930-adb2-caa138f0537f/jobs/868347) in the bip32 test-snap e2e test, is throw [here](https://github.com/MetaMask/key-tree/blob/3e87ae3f6e3803737bde5d51ef098c52be7c95f6/src/derivation.ts#L149) in `key-tree`. This is happening because the changed `bip39` implementation being pulled in in the latest `eth-keyring-controller` changes are causing the seedphrase to be passed as a `Uint8Array` rather than a plain text string as is expected in [validatePathSegment](https://github.com/MetaMask/key-tree/blob/3e87ae3f6e3803737bde5d51ef098c52be7c95f6/src/derivation.ts#L135).

Gudahtt (approved):

PR #107, opened 2023-01-17
Accept BIP-39 mnemonic phrase as Uint8Array

### LINKED ISSUE #106: Support mnemonics being passed as `Uint8Array`
Support mnemonics being passed as `Uint8Array` to let us get rid of passing mnemonics as string in the extension.

Mrtenz:
This adds support for passing mnemonic phrases as `Uint8Array`, in all places where we accept them. The expected format is the same as the format we use in `@metamask/scure-bip39`.

The derivation path can now look like `[new Uint8Array(), 'bip32:...', ...]`. Only the first BIP-39 segment can be a `Uint8Array`, it's not supported for BIP-32 segments.

Closes #106.

### DIFF
--- a/src/BIP44CoinTypeNode.test.ts
+++ b/src/BIP44CoinTypeNode.test.ts
@@ -229,26 +229,14 @@ describe('BIP44CoinTypeNode', () => {
         BIP44PurposeNodeToken,
         `bip32:60'`,
       ]);
-      const coinType = 60;
-      const pathString = `m / bip32:44' / bip32:${coinType}'`;
 
-      expect(node.coin_type).toStrictEqual(coinType);
-      expect(node.depth).toBe(2);
-      expect(node.privateKeyBytes).toHaveLength(32);
-      expect(node.publicKeyBytes).toHaveLength(65);
-      expect(node.path).toStrictEqual(pathString);
+      const stringNode = await BIP44CoinTypeNode.fromDerivationPath([
+        defaultBip39NodeToken,
+        BIP44PurposeNodeToken,
+        `bip32:60'`,
+      ]);
 
-      expect(node.toJSON()).toStrictEqual({
-        coin_type: coinType,
-        depth: node.depth,
-        masterFingerprint: node.masterFingerprint,
-        parentFingerprint: node.parentFingerprint,
-        index: node.index,
-        path: pathString,
-        privateKey: node.privateKey,
-        publicKey: node.publicKey,
-        chainCode: node.chainCode,
-      });
+      expect(node.toJSON()).toStrictEqual(stringNode.toJSON());
     });
 
     it('throws if derivation path has invalid depth', async () => {
--- a/src/BIP44CoinTypeNode.test.ts
+++ b/src/BIP44CoinTypeNode.test.ts
@@ -10,12 +10,10 @@ import {
 } from '.';
 import fixtures from '../test/fixtures';
 import { encodeExtendedKey, PRIVATE_KEY_VERSION } from './extended-keys';
-import { mnemonicPhraseToUint8Array } from './utils';
+import { mnemonicPhraseToBytes } from './utils';
 
 const defaultBip39NodeToken = `bip39:${fixtures.local.mnemonic}` as const;
-const defaultBip39BytesToken = mnemonicPhraseToUint8Array(
-  fixtures.local.mnemonic,
-);
+const defaultBip39BytesToken = mnemonicPhraseToBytes(fixtures.local.mnemonic);
 
 describe('BIP44CoinTypeNode', () => {
   describe('fromJSON', () => {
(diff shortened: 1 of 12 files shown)


src/BIP44CoinTypeNode.test.ts
```diff
@@ -221,6 +223,34 @@ describe('BIP44CoinTypeNode', () => {
       });
     });
 
+    it('initializes a BIP44CoinTypeNode with a Uint8Array', async () => {
```
FrederikBolding, line 226: Can we verify it against passing as a string?

src/BIP44Node.test.ts
```diff
@@ -122,6 +123,27 @@ describe('BIP44Node', () => {
       });
     });
 
+    it('initializes a new node from a derivation path with a Uint8Array', async () => {
```
FrederikBolding, line 126: Same here

src/SLIP10Node.test.ts
```diff
@@ -289,6 +290,29 @@ describe('SLIP10Node', () => {
       });
     });
 
+    it('initializes a new node from a derivation path with a Uint8Array', async () => {
+      const node = await SLIP10Node.fromDerivationPath({
```
FrederikBolding, line 294: Same here

src/derivation.ts
```diff
@@ -144,11 +154,17 @@ export function validatePathSegment(
   let startsWithBip39 = false;
   path.forEach((node, index) => {
     if (index === 0) {
-      startsWithBip39 = BIP_39_PATH_REGEX.test(node);
-      if (!startsWithBip39 && !BIP_32_PATH_REGEX.test(node)) {
+      startsWithBip39 =
+        node instanceof Uint8Array || BIP_39_PATH_REGEX.test(node);
+
+      if (
```
FrederikBolding, line 160: Please test the validation

src/utils.ts
```diff
@@ -329,3 +330,25 @@ export const getFingerprint = (publicKey: Uint8Array): number => {
...
+export function mnemonicPhraseToBytes(mnemonicPhrase: string): Uint8Array {
+  const words = mnemonicPhrase.split(' ');
+  const indices = words.map((word) => {
+    const index = englishWordlist.indexOf(word);
+    assert(index !== -1, `Invalid mnemonic phrase: Unknown word "${word}".`);
+
+    return index;
+  });
+
+  return new Uint8Array(new Uint16Array(indices).buffer);
```
FrederikBolding, line 353: Why is the `Uint8Array` created using a `Uint16Array`? 🤔

Mrtenz: The indices can be anything from 0 to 2047, which is too large to fit in an 8-bit number. By creating a `Uint16Array` and taking its underlying `ArrayBuffer`, we can easily split the 16-bit numbers into 8-bit numbers.

This is how it's done in `@metamask/scure-bip39` as well.

FrederikBolding: Interesting 👍

src/SLIP10Node.test.ts
```diff
@@ -289,6 +290,28 @@ describe('SLIP10Node', () => {
...
+        curve: 'secp256k1',
+      });
+
+      const stringNode = await SLIP10Node.fromDerivationPath({
+        derivationPath: [
+          defaultBip39NodeToken,
+          BIP44PurposeNodeToken,
+          `bip32:60'`,
+        ],
+        curve: 'secp256k1',
```
FrederikBolding, line 309: Should we test it for both curves?

src/derivation.test.ts
```diff
@@ -55,6 +55,30 @@ describe('derivation', () => {
...
+            ...ethereumBip32PathParts,
+            getUnhardenedBIP32NodeToken(index),
+          ] as const;
+
+          const multipath = [
+            mnemonicPhraseToBytes(mnemonic),
+            ...bip32Part,
+          ] as HDPathTuple;
+
+          return deriveKeyFromPath({ path: multipath, curve: 'secp256k1' });
```
FrederikBolding, line 71: Should we test for both curves?

src/derivation.ts
```diff
@@ -144,11 +154,17 @@ export function validatePathSegment(
...
+        !(node instanceof Uint8Array) &&
+        !startsWithBip39 &&
```
FrederikBolding, line 162: Wouldn't it be enough to check if `!startsWithBip39`?

src/SLIP10Node.test.ts
Mrtenz: Curve is a BIP-32 thing, not BIP-39. I can test it if you prefer, but I don't think it's necessary.

src/derivation.ts
Mrtenz: The `node instanceof Uint8Array` is necessary for TypeScript:

```
TS2345: Argument of type 'BIP39Node | BIP32Node' is not assignable to parameter of type 'string'.   Type 'Uint8Array' is not assignable to type 'string'.
```

FrederikBolding: Okay, can we leave a comment about that?

src/SLIP10Node.test.ts
FrederikBolding: Fair enough 👍

FrederikBolding (approved):

PR #118, opened 2023-02-28
Validate that master private key and seed are within bounds

Mrtenz:
This validates that:

- The master private key is greater than zero and less than the order of the curve.
- The seed is within 16 and 64 bytes, as per the BIP-32 specification.

Related to MM-02-003 and MM-02-004.


src/derivers/bip39.ts
```diff
@@ -47,11 +48,21 @@ export async function deriveChildKey({
  */
 export async function createBip39KeyFromSeed(
   seed: Uint8Array,
-  curve: Curve = secp256k1,
```
Mrtenz, line 50: I've made this non-optional, because we were forgetting to pass `ed25519` to this function in certain tests. It's better to be explicit here.

PR #120, opened 2023-03-02
Handle errors when resulting public or private key is invalid

Mrtenz:
This handles cases where the resulting public or private key after derivation is invalid (e.g., >= n, point at infinity, etc.). Previously we did not handle these situations at all. The chance of this happening is extremely low, but we need to handle it regardless to be compliant with BIP-32 and SLIP-10.

Unfortunately BIP-32 and SLIP-10 have a different way of handling this situation, so I had to refactor some code to make this possible:

- The `SLIP10Node` class now accepts a "specification" field (`bip32` or `slip10`), which determines how to handle errors.
   - This is always set to `bip32` for `BIP44Node`.
   - If no specification is provided, we default to `bip32` for `secp256k1`, and `slip10` for `ed25519`.
- Key derivation is now done sort of recursively, in that if the derivation fails, the same function is called again with the updated parameters (as described by the specification).


jest.config.js
```diff
@@ -33,7 +33,7 @@ module.exports = {
   // ],
 
   // Indicates which provider should be used to instrument code for coverage
-  coverageProvider: 'v8',
+  coverageProvider: 'babel',
```
Mrtenz, line 36: There were a few cases that couldn't be covered (both `assertExhaustive`), and `v8` doesn't support ignoring lines for coverage, as far as I know.

src/derivers/bip32.test.ts
```diff
@@ -2,16 +2,249 @@ import { bytesToHex, hexToBytes } from '@metamask/utils';
...
+      jest.spyOn(secp256k1, 'isValidPrivateKey').mockReturnValueOnce(false);
+
+      const childNode = await deriveChildKey({
+        node,
+        path: `0'`,
+        curve: secp256k1,
+        specification: 'bip32',
+      });
+
+      expect(childNode.index).toBe(BIP_32_HARDENED_OFFSET + 1);
```
Mrtenz, line 40: Currently this returns the index for the next node. This needs to be compared to reference implementations, as they may return the index that was originally being derived, even though `n + 1` is being derived under the hood.

src/derivers/bip32.test.ts
```diff
@@ -2,16 +2,249 @@ import { bytesToHex, hexToBytes } from '@metamask/utils';
...
   publicKeyToEthAddress,
 } from './bip32';
+import { bip39MnemonicToMultipath } from './bip39';
 
 const privateAddFixtures = fixtures['secp256k1-node'].privateAdd;
 
+describe('deriveChildKey', () => {
+  describe('private key derivation', () => {
+    it('handles invalid keys using BIP-32', async () => {
+      // TODO: Compare these results to reference implementations.
```
Mrtenz, line 22: I still have to compare all these values to reference implementations.

src/SLIP10Node.ts
```diff
@@ -64,6 +68,21 @@ export type JsonSLIP10Node = {
...
+   * While SLIP-10 and BIP-32 are largely compatible when the `secp256k1` curve
+   * is used, they differ in the way they handle key derivation errors. The
+   * probability of this happening is extremely low, but it is possible. When
+   * full compatibility with one of these specifications is required, this field
+   * should be set to the appropriate value.
+   *
+   * Note that `ed25519` is not supported by BIP-32, so the `bip32`
+   * specification is not available for this curve.
+   */
+  readonly specification: Specification;
```
Mrtenz, line 85: As alternative to the `specification` field, we could change the derivation paths to accept `slip10:${index}`, but that requires some refactoring.

src/derivers/bip32.ts
```diff
@@ -135,24 +149,293 @@ export async function deriveChildKey({
...
+          specification,
+        });
+      }
+
+      case 'slip10': {
+        // As per SLIP-10, if the resulting key is invalid, the new entropy is
+        // generated as follows:
+        // Key material (32 bytes), child chain code (32 bytes) =
+        //   HMAC-SHA512(parent chain code, 0x01 || chain code from invalid key || index).
+        const newEntropy = hmac(
```
FrederikBolding, line 409: Is there a reason not to use `generateEntropy` here?

FrederikBolding (approved):

PR #122, opened 2023-03-06
Allow zero private key for `ed25519`

Mrtenz:
`0x000...000` is a valid private key for `ed25519` (since all private keys are valid). Previously we would always reject this private key, due to the function used to validate bytes. I've updated it to use the curve function to check if a private key is valid.

Related to MM-02-007.

### DIFF
--- a/src/utils.test.ts
+++ b/src/utils.test.ts
@@ -22,6 +22,7 @@ import {
   encodeBase58check,
   decodeBase58check,
   mnemonicPhraseToBytes,
+  getBytesUnsafe,
 } from './utils';
 
 // Inputs used for testing non-negative integers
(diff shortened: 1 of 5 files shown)


src/utils.test.ts
```diff
@@ -308,6 +309,38 @@ describe('getBytes', () => {
...
+      'Invalid value: Must be a non-zero 1-byte byte array.',
+    );
+
+    expect(() => getBytes(new Uint8Array(1).fill(0), 1)).toThrow(
+      'Invalid value: Must be a non-zero 1-byte byte array.',
+    );
+  });
+});
+
+describe('getBytesUnsafe', () => {
```
FrederikBolding, line 330: Should we test that it doesn't throw if the buffer is zeroed?

FrederikBolding (approved):

PR #124, opened 2023-03-08
BREAKING: Remove specification field in favour of SLIP-10 path type

Mrtenz:
This replaces the specification field added in #120, with a new path type `slip10:`. This means that you can now use something like `["slip10:44'", "slip10:60'", /* ... */]` as derivation path, as well as the current `bip32:` derivation paths. This simplifies the BIP-32 derivation logic, and separates SLIP-10 and BIP-32 specific logic to a separate file.

While BIP-32 and SLIP-10 are mostly the same, there are some subtle differences. Projects that require full compatibility with one or the other spec can now choose based on the path type.

## Example

This works:

```ts
const node = SLIP10Node.fromExtendedKey({
  /* ... */,
  curve: 'secp256k1',
});

node.derive([`slip10:0'`, `slip10:1`]);
```

This also works:

```ts
const node = SLIP10Node.fromExtendedKey({
  /* ... */,
  curve: 'secp256k1',
});

node.derive([`bip32:0'`, `bip32:1`]);
```

This does **not** work anymore:

```ts
const node = SLIP10Node.fromExtendedKey({
  /* ... */,
  curve: 'ed25519', // <-- Note the curve
});

node.derive([`bip32:0'`, `bip32:1`]);
```

## Breaking changes

- `ed25519` can no longer be used with `bip32:` derivation paths, as it's only defined in SLIP-10.

### DIFF
--- a/src/derivers/shared.ts
+++ b/src/derivers/shared.ts
@@ -8,10 +8,33 @@ import { hmac } from '@noble/hashes/hmac';
 import { sha512 } from '@noble/hashes/sha512';
 
 import { DerivedKeys } from '.';
-import { BIP_32_HARDENED_OFFSET } from '../constants';
+import { BIP_32_HARDENED_OFFSET, UNPREFIXED_PATH_REGEX } from '../constants';
 import { Curve, mod } from '../curves';
 import { SLIP10Node } from '../SLIP10Node';
 import { isValidBytesKey, numberToUint32 } from '../utils';
+
+type BaseDeriveNodeArgs = {
+  entropy: Uint8Array;
+  chainCode: Uint8Array;
+  childIndex: number;
+  isHardened: boolean;
+  depth: number;
+  parentFingerprint: number;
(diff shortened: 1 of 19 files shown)


src/derivers/shared.ts
```diff
@@ -0,0 +1,540 @@
...
+ * @throws If the path is invalid.
+ */
+function validatePath(
+  path: string | Uint8Array,
+  node: SLIP10Node,
+  curve: Curve,
+): asserts path is string {
+  assert(typeof path === 'string', 'Invalid path: Must be a string.');
+
+  const isHardened = path.endsWith(`'`);
```
FrederikBolding, line 499: Shouldn't this be `includes()`?

Mrtenz: Hardened paths always end with `'`, which is enforced by our regex.

FrederikBolding (approved):

PR #125, opened 2023-03-08
Add randomly generated test vectors

Mrtenz:
This adds a script to generate test vectors (`yarn generate-vectors`), and a total of 800 different vectors, for BIP-32 and SLIP-10, `secp256k1` and `ed25519`.

Depends on #124.


scripts/generate-vectors.ts
```diff
@@ -0,0 +1,188 @@
...
+  const path = getRandomPath(spec, undefined, hardened);
+  const child = await node.derive(path);
+
+  return {
+    path: {
+      tuple: path,
+      string: path.join(' / '),
+    },
+    privateKey: child.privateKey,
+    publicKey: child.compressedPublicKey,
```
FrederikBolding, line 100: Is there a reason we don't use the full length public key?

Mrtenz: Not particularly. The compressed public key is smaller so there's less data in the vectors file, but either way it's not a big difference. From my understanding, compressed public keys are essentially just the `x`-coordinate on the curve, so if you want the point on the curve you have to calculate the `y`-coordinate first. Uncompressed public keys are a combination of `(x, y)`, so slightly faster to work with, as you don't need to calculate anything.

FrederikBolding (approved):

PR #131, opened 2023-06-19
Add `isValidBIP32PathSegment` function

Mrtenz:
This PR adds a new function, `isValidBIP32Path`, which can be used to check if a string is a valid BIP-32 path in the form `${number}'`. It will check if the index is `0 <= n <= 2**32 - 1`.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -28,7 +28,5 @@ export {
   getBIP44AddressKeyDeriver,
 } from './BIP44CoinTypeNode';
 export * from './constants';
-export type {
-  CoinTypeToAddressIndices,
-  isValidBIP32PathSegment,
-} from './utils';
+export type { CoinTypeToAddressIndices } from './utils';
+export { isValidBIP32PathSegment } from './utils';
--- a/src/index.ts
+++ b/src/index.ts
@@ -28,4 +28,7 @@ export {
   getBIP44AddressKeyDeriver,
 } from './BIP44CoinTypeNode';
 export * from './constants';
-export type { CoinTypeToAddressIndices, isValidBIP32Path } from './utils';
+export type {
+  CoinTypeToAddressIndices,
+  isValidBIP32PathSegment,
+} from './utils';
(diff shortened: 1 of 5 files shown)


src/index.ts
```diff
@@ -28,4 +28,7 @@ export {
   getBIP44AddressKeyDeriver,
 } from './BIP44CoinTypeNode';
 export * from './constants';
-export type { CoinTypeToAddressIndices } from './utils';
+export type {
```
FrederikBolding, line 31:
```suggestion
export {
```

src/utils.test.ts
```diff
@@ -197,6 +198,25 @@ describe('isValidBIP32Index', () => {
   });
 });
 
+describe('isValidBIP32PathSegment', () => {
+  it('returns true if the path is valid', () => {
```
FrederikBolding, line 202:
```suggestion
  it('returns true if the path segment is valid', () => {
```

src/utils.ts
```diff
@@ -170,6 +172,27 @@ export function isValidBIP32Index(index: number): boolean {
   return isValidInteger(index) && index <= MAX_BIP_32_INDEX;
 }
 
+/**
+ * Check if the value is a valid BIP-32 path segment, i.e., a string of the form
+ * `0'`.
+ *
+ * @param path - The BIP-32 path to test.
```
FrederikBolding, line 179:
```suggestion
 * @param segment - The BIP-32 path segment to test.
```
Maybe?

src/utils.ts
```diff
@@ -170,6 +172,27 @@ export function isValidBIP32Index(index: number): boolean {
...
+ * @returns Whether the path is a valid BIP-32 path.
```
FrederikBolding, line 180:
```suggestion
 * @returns Whether the segment is a valid BIP-32 path segment.
```

FrederikBolding (approved):

FrederikBolding (approved):

PR #155, opened 2023-10-08
BREAKING: Make all APIs synchronous

Mrtenz:
Blocked by #154.

All of the APIs we use from `@noble/curves` are synchronous, meaning that we can make all functions in this package synchronous as well. This is a breaking change, because it changes the function signature for most functions.


Mrtenz:
Synchronous APIs:

```
yarn test  61.28s user 3.85s system 212% cpu 30.655 total
```

Asynchronous APIs:

```
yarn test  75.17s user 4.44s system 245% cpu 32.381 total
```

PR #158, opened 2023-12-06
Support Cardano key derivation according to CIP3-Icarus

PeterBenc:
### What is the current state of things and why does it need to change?

- currently, this lib supports `secp256k1` and `ed25519`
- while developing a Snap for Cardano, we realized that keys derived by snap, using (`getBip32Entropy`) are different from keys that would standard Cardano wallet give, for the same mnemonic
- this inconsistency could lead to confusion and interoperability issues
- currently developed Cardano snap, relies on these changes, and is blocked by this. (note that changes need to be made also to `metamask-snaps` and `metamask-extension` repos, which are ready, awaiting feedback on this)

### What is the solution your changes offer and how does it work?

- by supporting derivation which is consistent with Cardano standard, we enable current Cardano users to use their existing mnemonics in Metamask and get the same wallet they used until now.
- similarily, we would enable current Metamask users to use their mnemonic in standard Cardano wallets and retrieve same addresses as they would through Cardano-Metamask snap.

For this reasons
- we add a new derivation `curve`, called `ed25519Bip32`, this curve implements a slight modification of how standard `ed25519` is used in SLIP10. 

- we add a new `deriver`, called `cip3Icarus` which derives keys according to `CIP3` (which is the standard for Cardano)

- from the interface point of view, these changes allow us to call 

  ```js
  await deriveKeyFromPath({
     path: ["cip3Icarus:1852'", "cip3Icarus: 1815'"],
     curve: 'ed25519Bip32',
  })
  ```

  which returns keys consistent with Cardano derivation standard

**A longer explanation, also present in the code**

[CIP-3](https://github.com/cardano-foundation/CIPs/blob/09d7d8ee1bd64f7e6b20b5a6cae088039dce00cb/CIP-0003/CIP-0003.md) defines standards for deriving keys on Cardano.
 
 Key attributes

   * Root/Master key is derived from entropy, not seed. For this implementation we work with **Icarus** standard as it is the most widely used. See https://github.com/cardano-foundation/CIPs/blob/09d7d8ee1bd64f7e6b20b5a6cae088039dce00cb/CIP-0003/Icarus.md.
     
   * HD node consists of a 64 byte private key, 32 byte public key and 32 byte chain code. See https://github.com/cardano-foundation/CIPs/blob/09d7d8ee1bd64f7e6b20b5a6cae088039dce00cb/CIP-0003/CIP-0003.md#master-key-generation.
     
   * For derivation of BIP32 HD nodes, it uses modified implementation of ed25519 called BIP32-Ed25519. See https://input-output-hk.github.io/adrestia/static/Ed25519_BIP.pdf.

### Are there any issues or other links reviewers should consult to understand this pull request better? For instance:

All sources are referenced in the code, and explained quite extensively, namely

- [CIP3](https://github.com/cardano-foundation/CIPs/blob/09d7d8ee1bd64f7e6b20b5a6cae088039dce00cb/CIP-0003/CIP-0003.md)
- [ed25519Bip32](https://input-output-hk.github.io/adrestia/static/Ed25519_BIP.pdf)

Reference implementations

- [cardano-js-sdk](https://github.com/input-output-hk/cardano-js-sdk/tree/master/packages/crypto)
- [cardano-crypto.js](https://github.com/vacuumlabs/cardano-crypto.js)

### DIFF
--- a/src/curves/ed25519Bip32.ts
+++ b/src/curves/ed25519Bip32.ts
@@ -77,16 +77,16 @@ export const getPublicKey = async (
 /**
  * Adds a tweak to a public key.
  *
- * @param _publicKey - The public key.
- * @param _tweak - The tweak to add.
+ * @param publicKey - The public key.
+ * @param tweak - The tweak to add.
  * @returns The resulting public key.
  */
 export const publicAdd = (
   publicKey: Uint8Array,
   tweak: Uint8Array,
 ): Uint8Array => {
-  return ed25519.ExtendedPoint.fromHex(remove0x(bytesToHex(_publicKey)))
-    .add(ed25519.ExtendedPoint.fromHex(remove0x(bytesToHex(_tweak))))
+  return ed25519.ExtendedPoint.fromHex(remove0x(bytesToHex(publicKey)))
+    .add(ed25519.ExtendedPoint.fromHex(remove0x(bytesToHex(tweak))))
     .toRawBytes();
 };
 
--- a/src/curves/ed25519Bip32.ts
+++ b/src/curves/ed25519Bip32.ts
@@ -82,4 +82,6 @@ export const decompressPublicKey = (publicKey: Uint8Array): Uint8Array => {
   return publicKey;
 };
 
+export const privateKeyLength = 64;
+
 export const masterNodeGenerationSpec = 'cip3Icarus';
--- a/src/curves/ed25519Bip32.ts
+++ b/src/curves/ed25519Bip32.ts
@@ -81,3 +81,5 @@ export const decompressPublicKey = (publicKey: Uint8Array): Uint8Array => {
   // Ed25519 public keys don't have a compressed form.
   return publicKey;
 };
+
+export const masterNodeGenerationSpec = 'cip3Icarus';
--- a/src/curves/ed25519Bip32.ts
+++ b/src/curves/ed25519Bip32.ts
@@ -1,0 +1,83 @@
+import {
+  stringToBytes,
+  bytesToHex,
+  hexToBigInt,
+  remove0x,
+} from '@metamask/utils';
+import { mod } from '@noble/curves/abstract/modular';
+import { ed25519 } from '@noble/curves/ed25519';
+
+export const curve = ed25519.CURVE;
+
+/**
+ * Named after whitepaper: BIP32-Ed25519 Hierarchical Deterministic Keys over a Non-linear Keyspace
+ * https://input-output-hk.github.io/adrestia/static/Ed25519_BIP.pdf
+ * "vanilla" "ed25519" curve follows SLIP10: https://tezos.stackexchange.com/questions/2837/can-i-use-bip32-hd-key-pairs-to-derive-ed25519-addresses
+ * note that that the important difference of the "bip32" version is that it allows unhardened key derivation
+ */
+export const name = 'ed25519Bip32';
+
+// Secret is empty string if not provided by user
+export const secret = stringToBytes('');
+
+/**
+ * Always returns true.
+ * For root node derivation, https://github.com/cardano-foundation/CIPs/blob/09d7d8ee1bd64f7e6b20b5a6cae088039dce00cb/CIP-0003/Icarus.md does not mention any cases when the private key is invalid.
+ * For child node derivation, https://input-output-hk.github.io/adrestia/static/Ed25519_BIP.pdf does not mention any cases when the private key is invalid.
+ *
+ * @param _privateKey - The private key to check.
+ * @returns True.
+ */
+export const isValidPrivateKey = (_privateKey: Uint8Array | string | bigint) =>
+  true;
+
+export const deriveUnhardenedKeys = true;
+
+export const publicKeyLength = 32;
+
+const bytesToNumberLE = (bytes: Uint8Array): bigint => {
+  return hexToBigInt(bytesToHex(Uint8Array.from(bytes).reverse()));
+};
+
+const modN = (a: bigint) => {
+  return mod(a, curve.n);
+};
+
+// Little-endian SHA512 with modulo n
+const modNLE = (hash: Uint8Array): bigint => {
+  return modN(bytesToNumberLE(hash));
+};
+
+// equivalent to https://github.com/jedisct1/libsodium/blob/93a6e79750a31bc0b946bf483b2ba1c77f9e94ce/src/libsodium/crypto_scalarmult/ed25519/ref10/scalarmult_ed25519_ref10.c#L105
+// which is used by cardano-js-sdk/crypto https://github.com/input-output-hk/cardano-js-sdk/blob/8a6db2a251cd1c956f52730a0d35de2b7fc67404/packages/crypto/src/Bip32/Bip32PrivateKey.ts#L161
+const multiplyWithBase = (key: Uint8Array): Uint8Array => {
+  const scalar = modNLE(key); // The actual scalar
+  const point = ed25519.ExtendedPoint.BASE.multiply(scalar); // Point on Edwards curve aka public key
+  return point.toRawBytes(); // Uint8Array representation
+};
+
+export const getPublicKey = async (
+  privateKey: Uint8Array,
+  _compressed?: boolean,
+): Promise<Uint8Array> => {
+  return multiplyWithBase(privateKey.slice(0, 32));
+};
+
+export const publicAdd = (
+  _publicKey: Uint8Array,
+  _tweak: Uint8Array,
+): Uint8Array => {
+  return ed25519.ExtendedPoint.fromHex(remove0x(bytesToHex(_publicKey)))
+    .add(ed25519.ExtendedPoint.fromHex(remove0x(bytesToHex(_tweak))))
+    .toRawBytes();
+};
+
+export const compressPublicKey = (publicKey: Uint8Array): Uint8Array => {
+  // Ed25519 public keys don't have a compressed form.
+  return publicKey;
+};
+
+export const decompressPublicKey = (publicKey: Uint8Array): Uint8Array => {
+  // Ed25519 public keys don't have a compressed form.
+  return publicKey;
+};
--- a/src/curves/ed25519Bip32.ts
+++ b/src/curves/ed25519Bip32.ts
@@ -39,19 +39,11 @@ const bytesToNumberLE = (bytes: Uint8Array): bigint => {
   return hexToBigInt(bytesToHex(Uint8Array.from(bytes).reverse()));
 };
 
-const modN = (a: bigint) => {
-  return mod(a, curve.n);
-};
-
-// Little-endian SHA512 with modulo n
-const modNLE = (hash: Uint8Array): bigint => {
-  return modN(bytesToNumberLE(hash));
-};
-
 // equivalent to https://github.com/jedisct1/libsodium/blob/93a6e79750a31bc0b946bf483b2ba1c77f9e94ce/src/libsodium/crypto_scalarmult/ed25519/ref10/scalarmult_ed25519_ref10.c#L105
 // which is used by cardano-js-sdk/crypto https://github.com/input-output-hk/cardano-js-sdk/blob/8a6db2a251cd1c956f52730a0d35de2b7fc67404/packages/crypto/src/Bip32/Bip32PrivateKey.ts#L161
 const multiplyWithBase = (key: Uint8Array): Uint8Array => {
-  const scalar = modNLE(key); // The actual scalar
+  // Little-endian SHA512 with modulo n
+  const scalar = mod(bytesToNumberLE(key), curve.n); // The actual scalar
   const point = ed25519.ExtendedPoint.BASE.multiply(scalar); // Point on Edwards curve aka public key
   return point.toRawBytes(); // Uint8Array representation
 };
--- a/src/curves/ed25519Bip32.ts
+++ b/src/curves/ed25519Bip32.ts
@@ -85,3 +85,5 @@ export const decompressPublicKey = (publicKey: Uint8Array): Uint8Array => {
 export const privateKeyLength = 64;
 
 export const masterNodeGenerationSpec = 'cip3';
+
+export const compressedPublicKeyLength = 32;
--- a/src/curves/ed25519Bip32.ts
+++ b/src/curves/ed25519Bip32.ts
@@ -41,7 +41,7 @@ export const publicKeyLength = 32;
  * @param bytes - The Uint8Array of bytes to convert.
  * @returns The converted bigint value.
  */
-const bytesToNumberLE = (bytes: Uint8Array): bigint => {
+export const bytesToNumberLE = (bytes: Uint8Array): bigint => {
   return hexToBigInt(bytesToHex(Uint8Array.from(bytes).reverse()));
 };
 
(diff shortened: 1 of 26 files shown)


src/utils.ts
```diff
@@ -438,13 +438,14 @@ export function validateCurve(
  * Get a 4-byte-long `Uint8Array` from a numeric value.
  *
  * @param value - The value to convert to a `Uint8Array`.
+ * @param littleEndian - Whether to use little endian byte order.
  * @returns The `Uint8Array` corresponding to the `bigint` value.
  */
-export function numberToUint32(value: number) {
+export function numberToUint32(value: number, littleEndian = false) {
```
PeterBenc, line 444: Needed later, to encode keyindex, for Cardano, the values are `littleEndian`

PeterBenc (commented):
Please review by commits, thank you :pray:

src/derivers/cip3Icarus.ts
Mrtenz, line 1: Sorry for the delay on this PR!

I'm not familiar with CIP-3 and the different key generation methods it describes. Is Icarus the main standard used by Cardano? If so, what do you think about renaming this to `cip3` instead of `cip3Icarus`? I think that simplifies the public API a little bit (`["cip3:0'", "cip3:..."]` versus `["cip3Icarus:0'", "cip3Icarus:..."]`). Icarus seems to be the recommended method by CIP-3, and if other methods need to be added in the future, we could add a suffix for those (like `cip3Trezor` etc.).

src/utils.ts
```diff
@@ -379,9 +379,9 @@ export const encodeBase58check = (value: Uint8Array): string => {
  * @returns The fingerprint of the public key.
  */
 export const getFingerprint = (publicKey: Uint8Array): number => {
-  if (!isValidBytesKey(publicKey, 33)) {
+  if (!isValidBytesKey(publicKey, 33) && !isValidBytesKey(publicKey, 32)) {
```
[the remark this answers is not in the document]
Mrtenz: I would prefer adding a new parameter, so we can keep the validation as strict as possible.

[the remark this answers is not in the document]
PeterBenc: Ok, I ll do that :+1:

src/derivers/cip3Icarus.ts
PeterBenc: Yes, `Icarus` is the main standard Cardano uses, but [Ledger](https://github.com/cardano-foundation/CIPs/blob/09d7d8ee1bd64f7e6b20b5a6cae088039dce00cb/CIP-0003/Ledger.md) also uses another significant one.
We were discussing this internally too, and initially, this was as you are proposing, just `cip3`. We then decided to be more specific and use `cip3Icarus`. When working on this feature, I found that the information is often not as straightforward as it should be, and bringing more clarity certainly does not hurt, even at the expense of the bit longer name.

Anyway, let me know what you think, I don't have a strong opinion about this, so we can change it to just `cip3` if you insist :+1:

Mrtenz: I feel like simplifying it in the API and providing good docs around this is better in this case.

PeterBenc: Ok, sure, will do that.

Mrtenz:
At first glance this looks great. I will be doing a proper review next week, thanks for your patience!

src/curves/ed25519Bip32.ts
```diff
@@ -0,0 +1,87 @@
...
+
+export const curve = ed25519.CURVE;
+
+/**
+ * Named after whitepaper: BIP32-Ed25519 Hierarchical Deterministic Keys over a Non-linear Keyspace
+ * https://input-output-hk.github.io/adrestia/static/Ed25519_BIP.pdf
+ * "vanilla" "ed25519" curve follows SLIP10: https://tezos.stackexchange.com/questions/2837/can-i-use-bip32-hd-key-pairs-to-derive-ed25519-addresses
+ * note that that the important difference of the "bip32" version is that it allows unhardened key derivation
+ */
+export const name = 'ed25519Bip32';
```
Mrtenz, line 18: Should we call this "bip32-ed25519" (since that's what the spec is called)?

src/curves/ed25519Bip32.ts
Mrtenz, line 1: Can you add JSDoc to all functions here?

src/derivers/bip39.ts
```diff
@@ -79,3 +86,49 @@ export async function createBip39KeyFromSeed(
...
+    c: 4096,
+    dkLen: 96,
+  });
+
+  // Consistent with the Icarus derivation scheme.
+  // https://github.com/cardano-foundation/CIPs/blob/09d7d8ee1bd64f7e6b20b5a6cae088039dce00cb/CIP-0003/Icarus.md
+  /* eslint-disable no-bitwise */
+  rootNode[0] &= 0b1111_1000;
+  rootNode[31] &= 0b0001_1111;
+  rootNode[31] |= 0b0100_0000;
```
Mrtenz, line 119:
```suggestion
  rootNode[31] |= 0b0100_0000;
  /* eslint-enable no-bitwise */
```

src/derivers/cip3Icarus.test.ts
```diff
@@ -0,0 +1,153 @@
...
+      const { bip39Node, purposeNode, coinTypeNode, accountNode, changeNode } =
+        fixture.nodes;
+      it('derives hardened private key 1', async () => {
+        const privateKey = await derivePrivateKey({
+          parentNode: fixtureNodeToParentNode(bip39Node),
+          isHardened: true,
+          childIndex: purposeNode.index - BIP_32_HARDENED_OFFSET,
+        });
+        expect(bytesToHex(privateKey)).toBe(purposeNode.privateKey);
+      });
```
Mrtenz:
```suggestion
      });

```

src/derivers/cip3Icarus.test.ts
```diff
@@ -0,0 +1,153 @@
...
+      it('derives hardened private key', async () => {
+        const privateKey = await derivePrivateKey({
+          parentNode: fixtureNodeToParentNode(purposeNode),
+          isHardened: true,
+          childIndex: coinTypeNode.index - BIP_32_HARDENED_OFFSET,
+        });
+        expect(bytesToHex(privateKey)).toBe(coinTypeNode.privateKey);
+      });
```
Mrtenz:
```suggestion
      });

```

src/derivers/cip3Icarus.test.ts
```diff
@@ -0,0 +1,153 @@
...
+    describe('deriveChainCode', () => {
+      const { bip39Node, purposeNode, accountNode, changeNode } = fixture.nodes;
+      it('derives hardened chainCode key', async () => {
+        const chainCode = await deriveChainCode({
+          parentNode: fixtureNodeToParentNode(bip39Node),
+          isHardened: true,
+          childIndex: purposeNode.index - BIP_32_HARDENED_OFFSET,
+        });
+        expect(bytesToHex(chainCode)).toBe(purposeNode.chainCode);
+      });
```
Mrtenz:
```suggestion
      });

```

src/derivers/cip3Icarus.test.ts
```diff
@@ -0,0 +1,153 @@
...
+        const purposeNodeRes = await deriveChildKey({
+          node: await SLIP10Node.fromJSON(bip39Node),
+          path: purpose,
+          curve: ed25519Bip32,
+        });
+
+        expect(JSON.stringify(purposeNodeRes)).toBe(
+          JSON.stringify(purposeNode),
+        );
+      });
```
Mrtenz:
```suggestion
      });

```

src/derivers/cip3Icarus.test.ts
```diff
@@ -0,0 +1,153 @@
...
+        const coinTypeNodeRes = await deriveChildKey({
+          node: await SLIP10Node.fromJSON(purposeNode),
+          path: coinType,
+          curve: ed25519Bip32,
+        });
+
+        expect(JSON.stringify(coinTypeNodeRes)).toBe(
+          JSON.stringify(coinTypeNode),
+        );
+      });
```
Mrtenz:
```suggestion
      });

```

src/derivers/cip3Icarus.ts
Mrtenz, line 1: Can you add JSDoc to the functions here?

Mrtenz (changes requested):
I didn't get through everything yet, but here's some initial comments for you. I'm not familiar with CIP-3, so I will take some time to read more about it to properly review the algorithms added here.

Some nits related to the code style we use. Sorry for that!

test/vectors/derivation.json
PeterBenc, line 1: Did this in separate commit so I can squash it easier

src/derivers/bip39.ts
```diff
@@ -34,16 +34,20 @@ export async function deriveChildKey({
   path,
   curve,
 }: DeriveChildKeyArgs): Promise<SLIP10Node> {
-  if (curve.masterNodeGenerationSpec === 'cip3') {
-    return entropyToCip3MasterNode(
-      mnemonicToEntropy(path, englishWordlist),
-      curve,
-    );
+  switch (curve.masterNodeGenerationSpec) {
+    case 'slip10':
```
PeterBenc, line 38: You haven't asked for this, but, I noticed it now, the switch is IMO safer

PeterBenc (commented):
@Mrtenz I was initially waiting for your full review, but then thought that you might be waiting for the `cip3Icarus` => `cip3` refactor, so I did it. Your review came literally one second after I pushed :D. Sorry for that, please check these 3 commits too.

src/utils.ts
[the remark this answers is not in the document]
PeterBenc: https://github.com/MetaMask/key-tree/pull/158/commits/4f546810aa59355c6692f1f2ef1b6802aa141e37

src/derivers/cip3Icarus.ts
PeterBenc: https://github.com/MetaMask/key-tree/pull/158/commits/a234957284d6637d99f27ffd014323c5e7e0c493

src/curves/ed25519Bip32.ts
```diff
@@ -0,0 +1,87 @@
...
+  return modN(bytesToNumberLE(hash));
+};
+
+// equivalent to https://github.com/jedisct1/libsodium/blob/93a6e79750a31bc0b946bf483b2ba1c77f9e94ce/src/libsodium/crypto_scalarmult/ed25519/ref10/scalarmult_ed25519_ref10.c#L105
+// which is used by cardano-js-sdk/crypto https://github.com/input-output-hk/cardano-js-sdk/blob/8a6db2a251cd1c956f52730a0d35de2b7fc67404/packages/crypto/src/Bip32/Bip32PrivateKey.ts#L161
+const multiplyWithBase = (key: Uint8Array): Uint8Array => {
+  const scalar = modNLE(key); // The actual scalar
+  const point = ed25519.ExtendedPoint.BASE.multiply(scalar); // Point on Edwards curve aka public key
+  return point.toRawBytes(); // Uint8Array representation
+};
```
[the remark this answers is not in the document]
PeterBenc: I inlined `modN` and `modNLE` which IMO makes sense, it was too simple to have in separate functions. Thanks :+1: 
I left `bytesToNumberLE` and `multiplyWithBase` in separate functions. Even though they are used only once, it makes sense to me (especially for `multiplyWithBase`) to be separate, since they are more complex, and own function is IMO a good space for documenting what these cryptic transformations do in plain English. 

Let me know what you think, naturally, I don't insist, i can inline them if you do :+1:

src/curves/ed25519Bip32.ts
PeterBenc: We had an internal discussion about this too. 

Regarding the `-`, we kept camel case, this is later imported with `import * as ed25519Bip32` which does not work with kebab case (afaik), and also is then put into object, which allows only camel case (we can disable eslint for that line ofc, but we decided that it's not worth it)

Initially, I called this `bip32Ed25519` but after discussions changed it to `ed25519Bip32` because 
> bip32 is "semantically" less important than the curve itself

Let me know what you think :)

src/derivers/cip3Icarus.ts
PeterBenc: https://github.com/MetaMask/key-tree/pull/158/commits/f0479a603475c0b26f3389f614d77addd8c5a674

PeterBenc:
@Mrtenz have you had time to look into `cip3`? 
PS: all relevant links are in the PR description :)

Mrtenz:
> @Mrtenz have you had time to look into `cip3`? PS: all relevant links are in the PR description :)

Not yet, sorry about that. Will make time to look into this more this week!

src/curves/ed25519Bip32.ts
```diff
@@ -0,0 +1,119 @@
...
+/**
+ * Adds a tweak to a public key.
+ *
+ * @param _publicKey - The public key.
+ * @param _tweak - The tweak to add.
+ * @returns The resulting public key.
+ */
+export const publicAdd = (
+  _publicKey: Uint8Array,
+  _tweak: Uint8Array,
```
Mrtenz, line 86:
```suggestion
  publicKey: Uint8Array,
  tweak: Uint8Array,
```

src/derivers/cip3.ts
```diff
@@ -0,0 +1,384 @@
...
+ * @param param1.childIndex - The index of the child key.
+ * @param param1.curve - Derivation curve.
+ * @returns The derived child public key.
+ */
+export const derivePublicKey = async ({
+  parentNode,
+  childIndex,
+  curve,
+}: DerivePublicKeyArgs) => {
+  // extension = (0x02||Ap||i)
```
Mrtenz, line 262: Just for clarification: The spec describes a way to derive public keys for `i >= 2^31`, but we don't need to implement that because we can use private key derivation in that case?

src/derivers/cip3.ts
```diff
@@ -0,0 +1,384 @@
...
+      depth: depth + 1,
+      parentFingerprint,
+      index: actualChildIndex,
+      curve: curve.name,
+    });
+  }
+
+  assert(
+    !isHardened,
+    "Cannot derive hardened because we don't have the private key",
```
Mrtenz, line 354:
```suggestion
    "Invalid parameters: Cannot derive hardened child keys without a private key.",
```

src/utils.ts
```diff
@@ -376,12 +376,16 @@ export const encodeBase58check = (value: Uint8Array): string => {
...
-export const getFingerprint = (publicKey: Uint8Array): number => {
-  if (!isValidBytesKey(publicKey, 33)) {
+export const getFingerprint = (
+  publicKey: Uint8Array,
+  compressedPublicKeyLength: number,
+): number => {
+  if (!isValidBytesKey(publicKey, compressedPublicKeyLength)) {
     throw new Error(
-      `Invalid public key: The key must be a 33-byte, non-zero byte array.`,
+      `Invalid public key: The key must be ${compressedPublicKeyLength} byte, non-zero byte array.`,
```
Mrtenz, line 388:
```suggestion
      `Invalid public key: The key must be a ${compressedPublicKeyLength}-byte, non-zero byte array.`,
```

Mrtenz (changes requested):
I reviewed the cryptography to the best of my ability and it looks good. Just a few more minor nits. Once they are fixed I will approve!

Mrtenz:
Can you rebase your branch?

PeterBenc:
@Mrtenz the last couple of issues should be resolved. Should I squash the commits or do you want to do it yourself?

Mrtenz:
`src/index.ts` is missing coverage(?), and you need to run `yarn lint:fix`. Looks good to me other than that!

PeterBenc:
Should be fixed

Should I squash the commits into some reasonable chunks, would that be relevant for audit too? do you want to do it yourself? Let me know :)

Mrtenz:
No need to. We want to do an audit for some other smaller changes since the last audit anyway, so I will just squash when merging this PR.

PR #175, opened 2024-05-15
Use `ts-bridge` for building

### LINKED ISSUE #152: bug: Invalid ESM build
The ESM build of the library should use the `.js` file extension for relative imports, else it is not a valid ES module.

Reference: https://nodejs.org/api/esm.html#mandatory-file-extensions

This also causes issues (MODULE_NOT_FOUND) with tools that support ESM (like Vitest).

Adding file extensions to relative imports should also cause no issues with the CJS build of the package.

Mrtenz:
This solves issues with the ESM build. Based on MetaMask/utils#182.

Closes #152.


tsconfig.json
```diff
@@ -3,7 +3,7 @@
     "esModuleInterop": true,
     "exactOptionalPropertyTypes": true,
     "forceConsistentCasingInFileNames": true,
-    "lib": ["ES2020"],
+    "lib": ["ES2020", "dom"],
```
FrederikBolding, line 6: Curious why we suddenly need this?

Mrtenz: Copied this from `utils` but doesn't seem necessary.

FrederikBolding (approved):

PR #196, opened 2024-11-22
Add extended key property to `SLIP10Node`

Mrtenz:
This adds an `extendedKey` property to SLIP-10 nodes, which returns the BIP-32 serialised extended key. This only works for `secp256k1` nodes, as other standards don't specify a standard for serialisation, and the length of the keys is different across curves.

I've also updated `SLIP10Node.fromExtendedKey` to accept a serialised extended key, instead of an options object.

### DIFF
--- a/src/BIP44Node.ts
+++ b/src/BIP44Node.ts
@@ -14,12 +14,7 @@ import {
   MIN_BIP_44_DEPTH,
 } from './constants';
 import type { SupportedCurve } from './curves';
-import {
-  decodeExtendedKey,
-  encodeExtendedKey,
-  PRIVATE_KEY_VERSION,
-  PUBLIC_KEY_VERSION,
-} from './extended-keys';
+import { decodeExtendedKey, PRIVATE_KEY_VERSION } from './extended-keys';
 import { SLIP10Node, validateBIP32Depth } from './SLIP10Node';
 import { isHardened } from './utils';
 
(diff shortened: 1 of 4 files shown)


src/BIP44Node.ts
```diff
@@ -161,24 +156,10 @@ export class BIP44Node implements BIP44NodeInterface {
...
+    validateBIP44Depth(options.depth);
 
     const node = await SLIP10Node.fromExtendedKey({
-      privateKey,
-      publicKey,
-      chainCode,
-      depth,
-      parentFingerprint,
-      index,
+      ...options,
```
FrederikBolding, line 162: What's the reasoning for being less explicit here now?

Mrtenz: Just seemed easier to read, but I can revert if you prefer.

src/SLIP10Node.ts
```diff
@@ -145,19 +163,86 @@ export class SLIP10Node implements SLIP10NodeInterface {
...
+      privateKey,
+      publicKey,
+      chainCode,
+      curve: curveName,
+    } = options;
+
     const chainCodeBytes = getBytes(chainCode, BYTES_KEY_LENGTH);
 
-    validateCurve(curve);
+    validateCurve(curveName);
```
FrederikBolding, line 245: Should we be asserting `curveName === 'secp256k1'` here as well?

FrederikBolding: I guess since we aren't decoding it isn't necessarily required, hmm

src/BIP44Node.ts
FrederikBolding: I think in this specific case its probably better to be explicit

src/SLIP10Node.test.ts
```diff
@@ -793,6 +895,76 @@ describe('SLIP10Node', () => {
...
+          const node = await SLIP10Node.fromExtendedKey({
+            privateKey,
+            chainCode,
+            curve: 'secp256k1',
+            depth: 0,
+            parentFingerprint: 0,
+            index: 0,
+          });
+
+          if (path.ours.tuple.length === 0) {
```
FrederikBolding, line 917: What is this meant to account for?

FrederikBolding: Ah, I had to remind myself of how the fixtures are structured

FrederikBolding (approved):

PR #197, opened 2024-11-22
Add cryptographic functions with option to specify custom functions

### LINKED ISSUE #195: Allow specifying custom hashing functions
For performance reasons, it may be beneficial to use custom hashing functions (i.e., `pbkdf2`, `hmac`) instead of the ones used in `key-tree` by default. Especially in React Native, the native function seems to be much quicker.

Mrtenz:
This adds cryptographic functions like `hmacSha512`, `pbkdf2Sha512`, with the option to use a custom implementation of these functions. This can be beneficial in environments like React Native, where the native implementation may be much faster than the JavaScript fallback.

These functions will now also use the Web Crypto API if available, which in itself results in a good performance improvement compared to the JavaScript functions.

Closes #195.

### DIFF
--- a/src/BIP44CoinTypeNode.ts
+++ b/src/BIP44CoinTypeNode.ts
@@ -10,6 +10,7 @@ import type {
   HardenedBIP32Node,
 } from './constants';
 import { BIP_32_HARDENED_OFFSET } from './constants';
+import type { CryptographicFunctions } from './cryptography';
 import type { SupportedCurve } from './curves';
 import { deriveChildNode } from './SLIP10Node';
 import type { CoinTypeToAddressIndices } from './utils';
@@ -75,19 +76,28 @@ export class BIP44CoinTypeNode implements BIP44CoinTypeNodeInterface {
    * @param json - The {@link JsonBIP44Node} for the key of this node.
    * @param coin_type - The coin_type index of this node. Must be a non-negative
    * integer.
+   * @param cryptographicFunctions - The cryptographic functions to use. If
+   * provided, these will be used instead of the built-in implementations.
    */
-  static async fromJSON(json: JsonBIP44Node, coin_type: number) {
+  static async fromJSON(
+    json: JsonBIP44Node,
+    coin_type: number,
+    cryptographicFunctions?: CryptographicFunctions,
+  ) {
     validateCoinType(coin_type);
     validateCoinTypeNodeDepth(json.depth);
 
-    const node = await BIP44Node.fromExtendedKey({
-      depth: json.depth,
-      index: json.index,
-      parentFingerprint: json.parentFingerprint,
-      chainCode: hexStringToBytes(json.chainCode),
-      privateKey: nullableHexStringToBytes(json.privateKey),
-      publicKey: hexStringToBytes(json.publicKey),
-    });
+    const node = await BIP44Node.fromExtendedKey(
+      {
+        depth: json.depth,
+        index: json.index,
+        parentFingerprint: json.parentFingerprint,
+        chainCode: hexStringToBytes(json.chainCode),
+        privateKey: nullableHexStringToBytes(json.privateKey),
+        publicKey: hexStringToBytes(json.publicKey),
+      },
+      cryptographicFunctions,
+    );
 
     return new BIP44CoinTypeNode(node, coin_type);
   }
@@ -107,13 +117,21 @@ export class BIP44CoinTypeNode implements BIP44CoinTypeNodeInterface {
    * `0 / 1 / 2 / 3 / 4 / 5`
    *
    * @param derivationPath - The derivation path for the key of this node.
+   * @param cryptographicFunctions - The cryptographic functions to use. If
+   * provided, these will be used instead of the built-in implementations.
    */
-  static async fromDerivationPath(derivationPath: CoinTypeHDPathTuple) {
+  static async fromDerivationPath(
+    derivationPath: CoinTypeHDPathTuple,
+    cryptographicFunctions?: CryptographicFunctions,
+  ) {
     validateCoinTypeNodeDepth(derivationPath.length - 1);
 
-    const node = await BIP44Node.fromDerivationPath({
-      derivationPath,
-    });
+    const node = await BIP44Node.fromDerivationPath(
+      {
+        derivationPath,
+      },
+      cryptographicFunctions,
+    );
 
     // Split the bip32 string token and extract the coin_type index.
     const pathPart = derivationPath[BIP_44_COIN_TYPE_DEPTH].split(
@@ -324,11 +342,14 @@ function validateCoinType(coin_type: unknown): asserts coin_type is number {
  * @param indices.account - The `account` index. Default: `0`.
  * @param indices.change - The `change` index. Default: `0`.
  * @param indices.address_index - The `address_index` index.
+ * @param cryptographicFunctions - The cryptographic functions to use. If
+ * provided, these will be used instead of the built-in implementations.
  * @returns The derived `address_index` key for the specified derivation path.
  */
 export async function deriveBIP44AddressKey(
   parentKeyOrNode: BIP44CoinTypeNode | JsonBIP44CoinTypeNode | string,
   { account = 0, change = 0, address_index }: CoinTypeToAddressIndices,
+  cryptographicFunctions?: CryptographicFunctions,
 ): Promise<BIP44Node> {
   const path = getBIP44CoinTypeToAddressPathTuple({
     account,
@@ -336,11 +357,14 @@ export async function deriveBIP44AddressKey(
     address_index,
   });
 
-  const node = await getNode(parentKeyOrNode);
-  const childNode = await deriveChildNode({
-    path,
-    node,
-  });
+  const node = await getNode(parentKeyOrNode, cryptographicFunctions);
+  const childNode = await deriveChildNode(
+    {
+      path,
+      node,
+    },
+    cryptographicFunctions,
+  );
 
   return new BIP44Node(childNode);
 }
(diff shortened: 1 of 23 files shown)


src/derivers/shared.ts
```diff
@@ -464,8 +463,11 @@ type GenerateEntropyArgs = {
  * @param args.extension - The extension bytes.
  * @returns The generated entropy bytes.
  */
-export function generateEntropy({ chainCode, extension }: GenerateEntropyArgs) {
-  return hmac(sha512, chainCode, extension);
+export async function generateEntropy({
```
FrederikBolding, line 466: Is this exported? If so, this is now breaking.

Mrtenz: It's not exported.

src/cryptography.ts
```diff
@@ -0,0 +1,176 @@
...
+  hmacSha512?: (key: Uint8Array, data: Uint8Array) => Promise<Uint8Array>;
+
+  /**
+   * Compute the PBKDF2 of the given password, salt, iterations, and key length.
+   * The hash function used is SHA-512.
+   *
+   * @param password - The password to hash.
+   * @param salt - The salt to use.
+   * @param iterations - The number of iterations.
+   * @param keyLength - The desired key length.
```
FrederikBolding, line 27:
```suggestion
   * @param keyLength - The desired key length in bytes.
```

src/cryptography.ts
```diff
@@ -0,0 +1,176 @@
...
+      password,
+      salt,
+      iterations,
+      keyLength,
+    );
+  }
+
+  if (isWebCryptoSupported()) {
+    /* eslint-disable no-restricted-globals */
+    const result = await crypto.subtle.importKey(
```
FrederikBolding, line 121: If possible we should test the two implementations against each other and compare the result

src/cryptography.ts
```diff
@@ -0,0 +1,176 @@
...
+  data: Uint8Array,
+  cryptographicFunctions: CryptographicFunctions = {},
+): Promise<Uint8Array> {
+  if (cryptographicFunctions.hmacSha512) {
+    return await cryptographicFunctions.hmacSha512(key, data);
+  }
+
+  if (isWebCryptoSupported()) {
+    /* eslint-disable no-restricted-globals */
+    const subtleKey = await crypto.subtle.importKey(
```
FrederikBolding, line 61: If possible we should test the two implementations against each other and compare the result

src/derivers/bip39.ts
```diff
@@ -1,17 +1,59 @@
...
+ *
+ * @param mnemonic - The mnemonic phrase to encode.
+ * @param wordlist - The wordlist to use.
+ * @returns The encoded mnemonic phrase.
+ */
+function encodeMnemonicPhrase(
+  mnemonic: string | Uint8Array,
+  wordlist: string[],
+) {
+  if (typeof mnemonic === 'string') {
```
FrederikBolding, line 26: In the scure implementation, this will throw if the mnemonic isn't of the right size

src/cryptography.ts
Mrtenz: There's a test that uses the Web Crypto API and one that doesn't, and we use the same inputs for both. Is that not sufficient?

src/cryptography.ts
FrederikBolding, line 1: With the introduction of this file, all of our existing tests would likely be using WebCrypto instead of the fallback implementation. Should we do something to test both code paths as extensively as we test everything else?

src/derivers/shared.ts
```diff
@@ -462,10 +500,15 @@ type GenerateEntropyArgs = {
...
  * @param args.chainCode - The parent chain code bytes.
  * @param args.extension - The extension bytes.
+ * @param cryptographicFunctions - The cryptographic functions to use. If
+ * provided, these will be used instead of the built-in implementations.
  * @returns The generated entropy bytes.
  */
-export function generateEntropy({ chainCode, extension }: GenerateEntropyArgs) {
-  return hmac(sha512, chainCode, extension);
+export async function generateEntropy(
+  { chainCode, extension }: GenerateEntropyArgs,
```
FrederikBolding, line 508: For non exported functions like this one, would it be cleaner to add `cryptographicFunctions` to the main args?

FrederikBolding: Might even be an option for exported functions if not breaking

Mrtenz: I prefer having it separately, as the main args are actually related to generating entropy, and the functions are just to pass them around to whichever function needs them.

src/derivers/bip39.ts
```diff
@@ -28,22 +75,27 @@ export function bip39MnemonicToMultipath(mnemonic: string): BIP39StringNode {
...
+  { path, curve }: DeriveChildKeyArgs,
+  cryptographicFunctions?: CryptographicFunctions,
+): Promise<SLIP10Node> {
   switch (curve.masterNodeGenerationSpec) {
     case 'slip10':
       return createBip39KeyFromSeed(
-        await mnemonicToSeed(path, englishWordlist),
+        await mnemonicToSeed(path, '', cryptographicFunctions),
         curve,
+        cryptographicFunctions,
```
FrederikBolding, line 91: General question for the entire PR, how much do we wanna test that `cryptographicFunctions` are passed around correctly?

src/BIP44CoinTypeNode.ts
```diff
@@ -75,19 +76,28 @@ export class BIP44CoinTypeNode implements BIP44CoinTypeNodeInterface {
...
    * @param coin_type - The coin_type index of this node. Must be a non-negative
    * integer.
+   * @param cryptographicFunctions - The cryptographic functions to use. If
+   * provided, these will be used instead of the built-in implementations.
    */
-  static async fromJSON(json: JsonBIP44Node, coin_type: number) {
+  static async fromJSON(
+    json: JsonBIP44Node,
+    coin_type: number,
+    cryptographicFunctions?: CryptographicFunctions,
```
FrederikBolding, line 85: Let's add similar tests to `SLIP10Node` that the functions are passed down correctly

src/cryptography.ts
```diff
@@ -0,0 +1,176 @@
...
+
+    const derivedBits = await crypto.subtle.deriveBits(
+      {
+        name: 'PBKDF2',
+        salt,
+        iterations,
+        hash: { name: 'SHA-512' },
+      },
+      result,
+      keyLength * 8,
```
FrederikBolding, line 137: Leave a comment explaining that this is to get the number of bits

src/cryptography.ts
```diff
@@ -0,0 +1,176 @@
...
+      password,
+      salt,
+      iterations,
+      keyLength,
+    );
+  }
+
+  if (isWebCryptoSupported()) {
+    /* eslint-disable no-restricted-globals */
+    const result = await crypto.subtle.importKey(
```
FrederikBolding, line 121:
```suggestion
    const key = await crypto.subtle.importKey(
```

src/derivers/bip39.ts
```diff
@@ -1,17 +1,91 @@
...
+  cryptographicFunctions?: CryptographicFunctions,
+) {
+  return await pbkdf2Sha512(
+    encodeMnemonicPhrase(mnemonic, englishWordlist),
+    stringToBytes(`mnemonic${passphrase}`.normalize('NFKD')),
+    2048,
+    64,
+    cryptographicFunctions,
+  );
+}
```
FrederikBolding, line 87:
```suggestion
  const salt = `mnemonic${passphrase}`.normalize('NFKD');
  return await pbkdf2Sha512(
    encodeMnemonicPhrase(mnemonic, englishWordlist),
    stringToBytes(salt),
    2048,
    64,
    cryptographicFunctions,
  );
}
```
Potentially worth it for readability

src/derivers/bip39.ts
```diff
@@ -28,22 +103,26 @@ export function bip39MnemonicToMultipath(mnemonic: string): BIP39StringNode {
...
     case 'slip10':
       return createBip39KeyFromSeed(
-        await mnemonicToSeed(path, englishWordlist),
+        await mnemonicToSeed(path, '', cryptographicFunctions),
         curve,
+        cryptographicFunctions,
       );
     case 'cip3':
       return entropyToCip3MasterNode(
         mnemonicToEntropy(path, englishWordlist),
```
FrederikBolding, line 123: Out of scope, but seems like we are really close to being able to drop `scure-bip39` as a dep if we implement this function

src/derivers/bip39.test.ts
```diff
@@ -4,16 +4,85 @@ import {
...
   deriveChildKey,
+  mnemonicToSeed,
 } from './bip39';
 
+const TEST_MNEMONIC_PHRASE =
+  'pill frown erosion humor invest inquiry rich garment seek such mention punch';
+
+describe('mnemonicToSeed', () => {
+  describe('without passphrase', () => {
+    const seed = new Uint8Array([
```
FrederikBolding, line 24: What's the source of truth for this seed that we are comparing with?

Mrtenz: https://github.com/MetaMask/scure-bip39/blob/612c6952ca8aee034e32dd0dc307c1d96e325ae2/test/bip39.test.ts#L127-L132

src/derivers/bip39.test.ts
```diff
@@ -4,16 +4,85 @@ import {
...
+        ),
+      ).rejects.toThrow(
+        'Invalid mnemonic phrase: The mnemonic phrase contains an unknown word.',
+      );
+    });
+  });
+
+  describe('with passphrase', () => {
+    const passphrase = 'passphrase';
+    const seed = new Uint8Array([
```
FrederikBolding, line 62: Same question applies here

src/derivers/bip39.test.ts
FrederikBolding: Let's reference that as a comment for our future selves 😄

FrederikBolding (approved):

PR #200, opened 2024-11-26
Add network property

Mrtenz:
BIP-32 specifies four different "versions" for extended keys. Public and private, but also mainnet and testnet. Previously we just assumed mainnet, because that's what's used for Ethereum, but to support other networks (such as Bitcoin), I've added a network field (i.e., `mainnet` or `testnet`), which determines which BIP-32 extended key version to use.

### DIFF
--- a/src/SLIP10Node.ts
+++ b/src/SLIP10Node.ts
@@ -2,19 +2,18 @@ import { assert, bytesToHex } from '@metamask/utils';
 
 import type { BIP44CoinTypeNode } from './BIP44CoinTypeNode';
 import type { BIP44Node } from './BIP44Node';
-import type { RootedSLIP10PathTuple, SLIP10PathTuple } from './constants';
+import type {
+  Network,
+  RootedSLIP10PathTuple,
+  SLIP10PathTuple,
+} from './constants';
 import { BYTES_KEY_LENGTH } from './constants';
 import type { CryptographicFunctions } from './cryptography';
 import type { SupportedCurve } from './curves';
 import { getCurveByName } from './curves';
 import { deriveKeyFromPath } from './derivation';
 import { publicKeyToEthAddress } from './derivers/bip32';
-import {
-  decodeExtendedKey,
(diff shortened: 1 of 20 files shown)


src/extended-keys.ts
```diff
@@ -1,41 +1,97 @@
...
   depth: number;
   parentFingerprint: number;
   index: number;
+  network: Network;
   chainCode: Uint8Array;
 };
 
 type ExtendedPublicKey = ExtendedKeyLike & {
-  version: typeof PUBLIC_KEY_VERSION;
+  type: 'public';
```
FrederikBolding, line 28: Is it intentional for this to be a breaking change? Could we not include the version bytes still?

src/SLIP10Node.ts
```diff
@@ -352,12 +365,14 @@ export class SLIP10Node implements SLIP10NodeInterface {
    * @param options.derivationPath - The rooted HD tree path that will be used
    * to derive the key of this node.
    * @param options.curve - The curve used by the node.
+   * @param options.network - The network for the node. This is only used for
+   * extended keys, and defaults to `mainnet`.
```
FrederikBolding, line 369: This is breaking too if we were assuming testnet before?

Mrtenz: Made a typo in the PR description. Ethereum uses `mainnet`.

test/fixtures.ts
```diff
@@ -458,6 +458,7 @@ export default {
           masterFingerprint: 1247889873,
           parentFingerprint: 0,
           index: 0,
+          network: 'mainnet',
```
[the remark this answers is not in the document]
Mrtenz: It's because of how the tests using these are written. They do a direct comparison of the JSON objects.

src/extended-keys.ts
Mrtenz: This is not a public type, so this is not a breaking change.

FrederikBolding (approved):

Issue #210, opened 2025-03-06
Support specifying the master seed instead of the raw mnemonic bytes for root nodes


FrederikBolding:
For performance reasons we should look into allowing passing the master seed instead of the raw mnemonic bytes for root nodes. For most curves supported by key-tree, the MetaMask clients already have this seed computed and we would be able to skip the most expensive iteration when deriving a node.

PR #212, opened 2025-03-06
Add `fromSeed` function to `SLIP10Node`, `BIP44Node`, `BIP44CoinTypeNode`

Mrtenz:
This adds a new function to initialise a `key-tree` node from a BIP-39 seed. This can be used to skip the step from mnemonic phrase to seed, and can improve performance if the seed is already available.

This function is only supported for `secp256k1` and `ed25519` for now. `ed25519Bip32` may be added in the future.

Closes https://github.com/MetaMask/key-tree/issues/210

### DIFF
--- a/src/BIP44CoinTypeNode.test.ts
+++ b/src/BIP44CoinTypeNode.test.ts
@@ -333,6 +333,74 @@ describe('BIP44CoinTypeNode', () => {
     });
   });
 
+  describe('fromSeed', () => {
+    it('initializes a BIP44CoinTypeNode from a seed', async () => {
+      const node = await BIP44CoinTypeNode.fromSeed({
+        derivationPath: [
+          fixtures.local.seed,
+          BIP44PurposeNodeToken,
+          `bip32:60'`,
+        ],
+      });
+
+      const mnemonicNode = await BIP44CoinTypeNode.fromDerivationPath([
+        defaultBip39NodeToken,
+        BIP44PurposeNodeToken,
+        `bip32:60'`,
+      ]);
+
+      expect(node.toJSON()).toStrictEqual(mnemonicNode.toJSON());
+    });
+
+    it('initializes a BIP44CoinTypeNode from a seed with custom cryptographic functions', async () => {
+      const functions = getMockFunctions();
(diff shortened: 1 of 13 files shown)


src/derivers/bip39.ts
```diff
@@ -114,17 +186,22 @@ export async function deriveChildKey(
   { path, curve, network }: DeriveChildKeyArgs,
   cryptographicFunctions?: CryptographicFunctions,
 ): Promise<SLIP10Node> {
+  assert(
+    path instanceof Uint8Array,
```
FrederikBolding, line 190: Is `path` always the seed now? Should we rename?

src/BIP44CoinTypeNode.test.ts
```diff
@@ -333,6 +333,74 @@ describe('BIP44CoinTypeNode', () => {
...
+      const pathString = `m / bip32:44' / bip32:${coinType}'`;
+
+      expect(node.coin_type).toStrictEqual(coinType);
+      expect(node.depth).toBe(2);
+      expect(node.privateKeyBytes).toHaveLength(32);
+      expect(node.publicKeyBytes).toHaveLength(65);
+      expect(node.path).toStrictEqual(pathString);
+
+      expect(functions.hmacSha512).toHaveBeenCalledTimes(3);
+      expect(functions.pbkdf2Sha512).not.toHaveBeenCalled();
```
FrederikBolding, line 378: 📈

src/derivers/bip39.ts
Mrtenz: It is. We can't rename it though, since this is based on a standard interface used for BIP-32, BIP-39, CIP-3.

FrederikBolding: Oh I guess we may not want to rename because the other derives implement `deriveChildKey` as a shared interface?

Maybe we should just update the doc string or somehow make it clearer that this has somewhat different behavior

FrederikBolding: Oh wow I should have refreshed😄

FrederikBolding: Can we add a note about it at least?

FrederikBolding (approved):

PR #213, opened 2025-03-07
Optimise performance of getting public keys

### LINKED ISSUE #211: Re-use public key computations more when deriving
There's a couple of instances of code paths where we are recomputing the public key when we have already computed it one scope up.

One instance of this is [here](https://github.com/MetaMask/key-tree/blob/0a2faf3aacf931876bd18c61055cde715606f76b/src/derivers/bip39.ts#L170) and [here](https://github.com/MetaMask/key-tree/blob/main/src/SLIP10Node.ts#L315), but I believe there are others too.

 This is worth revisiting for performance reasons.

Mrtenz:
This optimises the performance of getting public keys in a couple of different ways:

- The `Curve.getPublicKey` function was changed to always return the public key synchronously. Every curve was already doing public key computations synchronously, so this lets us remove redundant `await`s and async functions.
- In cases where we need to compute the public key (to get the fingerprint for example), that public key is now re-used when possible, to avoid computing it twice.
- If a public key is not provided to `SLIP10Node`, it's now lazily computed, and cached. This avoids computing it unnecessarily when it's not used.

Closes #211.

### DIFF
--- a/src/SLIP10Node.ts
+++ b/src/SLIP10Node.ts
@@ -2,12 +2,12 @@ import { assert, bytesToHex } from '@metamask/utils';
 
 import type { BIP44CoinTypeNode } from './BIP44CoinTypeNode';
 import type { BIP44Node } from './BIP44Node';
+import { BYTES_KEY_LENGTH, PUBLIC_KEY_GUARD } from './constants';
 import type {
   Network,
   RootedSLIP10PathTuple,
   SLIP10PathTuple,
 } from './constants';
-import { BYTES_KEY_LENGTH } from './constants';
 import type { CryptographicFunctions } from './cryptography';
 import type { SupportedCurve } from './curves';
 import { getCurveByName } from './curves';
@@ -99,17 +99,31 @@ export type SLIP10NodeInterface = JsonSLIP10Node & {
   toJSON(): JsonSLIP10Node;
 };
 
-export type SLIP10NodeConstructorOptions = {
+type BaseSLIP10NodeConstructorOptions = {
   readonly depth: number;
   readonly masterFingerprint?: number | undefined;
   readonly parentFingerprint: number;
   readonly index: number;
   readonly network?: Network | undefined;
   readonly chainCode: Uint8Array;
-  readonly privateKey?: Uint8Array | undefined;
-  readonly publicKey: Uint8Array;
   readonly curve: SupportedCurve;
 };
+
+type SLIP10NodePrivateKeyConstructorOptions =
+  BaseSLIP10NodeConstructorOptions & {
+    readonly privateKey: Uint8Array;
+    readonly publicKey?: Uint8Array | undefined;
+  };
+
+type SLIP10NodePublicKeyConstructorOptions =
+  BaseSLIP10NodeConstructorOptions & {
+    readonly privateKey?: Uint8Array | undefined;
+    readonly publicKey: Uint8Array;
+  };
+
+export type SLIP10NodeConstructorOptions =
+  | SLIP10NodePrivateKeyConstructorOptions
+  | SLIP10NodePublicKeyConstructorOptions;
 
 export type SLIP10ExtendedKeyOptions = {
   readonly depth: number;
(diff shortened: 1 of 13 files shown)


src/SLIP10Node.ts
```diff
@@ -503,15 +530,20 @@ export class SLIP10Node implements SLIP10NodeInterface {
       'SLIP10Node can only be constructed using `SLIP10Node.fromJSON`, `SLIP10Node.fromExtendedKey`, `SLIP10Node.fromDerivationPath`, or `SLIP10Node.fromSeed`.',
     );
 
+    assert(
+      privateKey !== undefined || publicKey !== undefined,
```
FrederikBolding, line 534: How can this happen now?

Mrtenz: I don't think it can, but this is just for sanity.

src/SLIP10Node.ts
```diff
@@ -299,11 +320,17 @@ export class SLIP10Node implements SLIP10NodeInterface {
...
       );
+
       assert(
         curveObject.isValidPrivateKey(privateKeyBytes),
         `Invalid private key: Value is not a valid ${curve} private key.`,
       );
 
+      const trustedPublicKey =
+        guard === PUBLIC_KEY_GUARD && publicKey
+          ? getBytes(publicKey, curveObject.publicKeyLength)
```
FrederikBolding, line 331: How come the passed `publicKey` isn't usable by itself?

src/constants.ts
```diff
@@ -6,6 +6,14 @@ export const MAX_BIP_44_DEPTH = 5;
 export const MAX_UNHARDENED_BIP_32_INDEX = 0x7fffffff; // 2^31 - 1
 export const MAX_BIP_32_INDEX = 0xffffffff; // 2^32 - 1
 
+/**
+ * A guard symbol to prevent untrusted public keys from being passed to
+ * `SLIP10Node` constructors.
+ */
+export const PUBLIC_KEY_GUARD = Symbol(
```
FrederikBolding, line 13: Right now this is exported in the index 😄

src/SLIP10Node.ts
Mrtenz: `publicKey` is typed as `string | Uint8Array`. `getBytes` will just return the `Uint8Array` as-is in this case.

FrederikBolding: So likely this does nothing, assuming we always pass in `publicKey` as `Uint8Array` internally?

Mrtenz: Yep.

FrederikBolding: Leave a comment please!

src/constants.ts
FrederikBolding: https://github.com/MetaMask/key-tree/blob/cc54abfd15b47abaa9feafaf3bd6488da306bebe/src/index.ts#L31

Mrtenz: I wrongfully assumed we only did explicit exports instead of wildcard exports. Will move the guard to a separate file.

FrederikBolding: Probably worth testing in the index.test that it isn't exported

Mrtenz: Moved to `guard.ts` and added a test to ensure nothing from `guard.ts` is exported.

src/derivers/shared.ts
```diff
@@ -223,13 +227,15 @@ async function deriveNode(
  *
  * @param options - The options for deriving a secret extension.
  * @param options.privateKey - The parent private key bytes.
+ * @param options.publicKey - The parent public key bytes.
```
FrederikBolding, line 230: Probably worth documenting that this is compressed public key

Mrtenz: Would've prevented some debugging 😅 That's why I added the check as well.

src/derivers/bip39.ts
```diff
@@ -311,7 +315,7 @@ export async function entropyToCip3MasterNode(
   assert(curve.isValidPrivateKey(privateKey), 'Invalid private key.');
 
   const masterFingerprint = getFingerprint(
-    await curve.getPublicKey(privateKey),
+    curve.getPublicKey(privateKey),
     curve.compressedPublicKeyLength,
   );
```
FrederikBolding, line 321: Can we not pass in the pubkey here too?

Mrtenz: Missed this one, updated!

FrederikBolding (approved):

PR #220, opened 2025-09-05
feat: add uint8 array compare function

hmalik88:
Adding a compare function for `Uint8` arrays. This compare function can be useful in various packages e.g. comparing mnemonics.


hmalik88:
Moving to @metamask/utils

